PR #6, opened 2024-02-20
New syntax

uncomputable:


doc/new-syntax.md
```diff
@@ -0,0 +1,337 @@
+# Goal
+
+Simphony code is valid Rust code _(assuming that used types, functions and constants are imported)_.
+
+Writing sum values as `Either<A, B>` instead of `A + B` might seem pointless or verbose.
```
sanket1729, line 5: +1

sanket1729:
I like the direction this is going towards. What does the [dosn't work] mean in section titles, do other programs compile/translate with the current rust-simplicity master?

doc/new-syntax.md
```diff
@@ -0,0 +1,337 @@
...
+| `[A; 1]`       | `Option<A>`           | List of length ≤ 1           |
+| `[A; N + 1]`   | `Option<(A, [A; N])>` | List of length ≤ N + 1       |
+
+The Boolean type is essentially a macro for a sum type of two units.
+
+The unsigned integer types are essentially macros for product types.
+
+The list type `[A; N]` can hold up to `N` values of type `A` where `N` is a positive integer.
+
+We need to know the list size at commit time, so `N` will be somewhat large. There will likely be parts of the list that are unpopulated at redeem time. The definition of `[A; N]` as nested option-pairs keeps the onchain footprint small, since the unpopulated parts don't contribute extra bits to the onchain encoding.
```
apoelstra, line 37: Honestly I'd be tempted to use the syntax `List<A, N>` rather than `[A; N]` to emphasize that this is something weird and not very array-like. In particular it has linear access time.

doc/new-syntax.md
```diff
@@ -0,0 +1,337 @@
...
+
+# Variables
+
+Expression `v` where `v` is a variable name.
+
+# Jets
+
+Expression `jet_j` where `j` is the name of a jet with source type `()`.
+
+Expression `jet_j e` where `j` is the name of a jet with source type other than `()`, and `e` is an expression.
```
apoelstra, line 63: Should we use `jet_j(e)` to make Rust programmers more comfortable?

apoelstra: Yeah, I think that's reasonable.

uncomputable:
Let's finalize a minimum-viable Simphony syntax + semantics without loops. The documentation is pretty much finished. Feel free to comment. I will start modifying the code base.

One open question is what `{scoping}` does. The syntax has rules like `e1; e2`. Are the `{brackets}` required to show precedence of operations, as in `(1 + 2) * 3`?

apoelstra:
@uncomputable I think scoping should affect the `let` statement, which assigns name(s) to expressions. Any names that appear within { and } should be invisible outside of them.

Similarly these names should be invisible *before* the let statement.

apoelstra:
We probably need to tweak the context in some subtle way to handle this (maybe just having a stack of contexts where } pops everything off since the last {?

These braces shouldn't have anything to do with precedence -- only ( ) parens should.

PR #8, opened 2024-03-07
New grammar

uncomputable:
Implementation of new syntax and semantics from #6: types, values, match statements, unwraps.

The Simplicity compiler panics on every error, as on master. I will fix this in a follow-up PR.

Types and patterns are parsed nonrecursively. Expressions are more complicated and are currently parsed recursively. There will be a follow-up PR.


uncomputable:
@apoelstra I don't expect an in-depth review, but feel free to skim the code and tell me what you see.

Issue #9, opened 2024-03-15
Test example programs with witness data


uncomputable:
Currently we test only a handful of programs a constant `20` for every witness node. We should write witness files for each program and include all programs in the unit test.

uncomputable:
Resolved in #63

PR #10, opened 2024-03-15
Update examples

uncomputable:
Use the new syntax and single input/output jets.

The examples will be tested once we can handle witness data (#9) and once the jets are defined (https://github.com/BlockstreamResearch/simplicity/pull/224).


apoelstra:
Nice!

utACK.

I guess the extensions on all these ought to be .simf rather than .simpl.

uncomputable:
The utACK is missing a git revision, but I will merge anyway.

PR #13, opened 2024-03-20
Array

uncomputable:
Add array type + expression + example.


apoelstra:
e38b11393d9ce32fa98cbaa0349d780f24366099 looks good. It's a bit of a surprising restriction that array sizes can only be powers of 2. I wonder if we should change the grammar to require the number be encoded as `2^n` rather than `n`-with-restrictions.

Or are you planning to extend the type to handle arbitrary widths?

Also, is the syntax where you assign numeric literals to e.g. `[bool; 128]` a special case for boolean arrays?

uncomputable:
Restricting array sizes to powers of two will confuse people, good point. Right now, arrays are a macro for nested products, but we might want to extend the definition to some sort of balanced binary tree of elements.

Assigning numeric literals to types such as `[bool; 128]` is in line with one of Simfony's design principles: Macro types behave exactly like the underlying base type. The base type of `[bool; 128]` and of `[u128]` is the same, so the same values can be assigned to both.

apoelstra:
> The base type of `[bool; 128]` and of `u128` is the same, so the same values can be assigned to both

Sounds good. We should make sure that's clear in the docs.

uncomputable:
Updated array types to be balanced trees of product types.

If an array has `n` elements, then we go to the next power of two `m ≥ n`. We split the array into `m / 2` elements and `n - m / 2` elements. And so on.

apoelstra:
ucACK cf49a5cbf8dac8fd2b2cba2e06f475e3626dbce3

At some point soon I should set up a test rig for this repo.

I'd kinda like a unit test for the `BTreeSlice` type. I think it'd be easy to create a few trees of different sizes with consecutive integers and check that you can `fold` them down to their sum, for example.

apoelstra (approved):
ACK 9b12ffff84941c4c65533085a1179cadcb56f041

PR #17, opened 2024-04-03
Array patterns

uncomputable:
Enable destructing of arrays in let statements `let [a, b,c] = (1, 2, 3)`, using array patterns. Do some refactors beforehand.

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -1,8 +1,7 @@
+use miniscript::iter::{Tree, TreeLike};
 use std::fmt;
 use std::ops::Range;
 use std::sync::Arc;
-
-use miniscript::iter::{Tree, TreeLike};
 
 /// View of a slice as a balanced binary tree.
 /// The slice must be nonempty.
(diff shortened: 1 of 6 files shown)


src/array.rs
```diff
@@ -183,6 +187,125 @@ impl<'a, A: Clone> TreeLike for Partition<'a, A> {
...
+///    / \
+///   .   c
+///  / \
+/// a   b
+/// ```
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum BinaryTree<T> {
+    /// Normal node (with at most two children).
+    ///
+    /// Must be constructed via [`BinaryTree::from_tree()`].
```
apoelstra, line 220: In 61d8b2393ea2d0690e11e38c2b10fcd58b9e1efe:

This is fine, but if you want to say "must be constructed via X" you really ought to split the type into `BinaryTreeInner` (a non-public enum) and `BinaryTree` (a public struct with a private `inner` field). Then callers literally cannot construct the type except the way you say so.

src/array.rs
```diff
@@ -183,6 +187,125 @@ impl<'a, A: Clone> TreeLike for Partition<'a, A> {
...
+                    let range = 0..array.len();
+                    Self::from_array(array, range).as_node()
+                }
+            },
+            Self::Array(array, range) => {
+                debug_assert!(2 <= range.len());
+                let n = range.len();
+                let next_pow2 = n.next_power_of_two();
+                let half = next_pow2 / 2;
+                let left = Self::from_array(array.clone(), range.start..range.start + half);
```
apoelstra, line 291: In 61d8b2393ea2d0690e11e38c2b10fcd58b9e1efe:

as a style nit, I prefer calling `Arc::clone` explicitly rather than `.clone()` so it's obvious that we're just doing a refcount increment and not a "real" clone.

src/array.rs
```diff
@@ -306,6 +305,85 @@ impl<T: TreeLike + fmt::Display> fmt::Display for BinaryTree<T> {
     }
 }
 
+/// One step on a path from ancestor nodes to descendent nodes.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub enum Direction {
```
apoelstra, line 310: In d1d5c78388208d4e16a60a2059044b42c01ba7e0:

I believe this enum should not be `pub` because it's an implementation detail of `DirectedTree`.

apoelstra: Ahh, no, it is returned from `find`. Interesting.

I wonder if you want to encapsulate the returned `Vec<Direction>` into a `DirectedTreePath` or something which hides the `Direction` type. Would result in a simpler API and be easier to reason about where direction-vectors come from.

Sadly in Rust you can't tie the `DirectedTreePath` to a specific directed tree the way you could with a dependent type system. (You could put a reference to the `DirectedTree` into the path but that's super unergonomic because it prevents moving the tree. And you could do some crazy run-time "tree ID" thing, maybe gated to only compile in debug mode, but that's a lot of work and shouldn't be in this PR.)

apoelstra:
As a terminology note. I think we should use the term "structural types" rather than "base types" or "Simplicity base types". This is easier to understand because it's a more unique name and describes exactly what they are (types which are defined by their structure, i.e. their pattern of sums and products).

In future we probably want a tighter distinction between these, so that it's no longer possible to e.g. assign an array literal to a manually-constructed tree of products, and so that it's possible for users to define types like `Celsius` and `Farenheit` and Simfony will prevent assigning between them. We could introduce a `cast` operator which would check that two types were structurally equal and refuse to compile otherwise.

But this may be out of scope for 1.0. (Although within 1.0, we may want to introduce more rigor into this codebase by distinguishing between Simfony types and inserting casts when they are needed.) For now let's at least use the "structural type" terminology though.

>

apoelstra:
This comment mostly applies to 4a10d48b767c09a53243e12fcde7964eb5ff3c84 "Pattern: convert to base".

apoelstra:
Done reviewing f3d4d2534fcd651923938af49f22441841354a90. Will ACK though I leave it up to you whether you want to address my comments here or in a followup PR.

apoelstra (approved):
ACK f3d4d2534fcd651923938af49f22441841354a90

src/array.rs
uncomputable: I left the comment "must be constructed via X" as a note to self, because I violated loop invariants by manually constructing structs. It is even better to enforce loop invariants via the compiler, which I will do. Thanks for pointing this out.

src/parse.rs
```diff
@@ -40,6 +40,39 @@ pub enum Pattern {
     Product(Arc<Self>, Arc<Self>),
 }
 
+impl Pattern {
+    /// Create an equivalent pattern that corresponds to the Simplicity base types.
```
uncomputable, line 56: 3d65d4b341b628a4d46dfb280b29551be1ca9a92: I'm not sure if "structural types" would be a better name. Arguably, values of a type always follow the type's structure.

We are trying to describe the types that are part of the definition of the Simplicity type system. These types are not defined in terms of other types.

uncomputable:
For now I updated `BinaryTree` to include a private enum.

src/parse.rs
uncomputable: Ideally we choose a name that fits both the Simplicity side and the Simfony side.

src/array.rs
```diff
@@ -183,6 +187,150 @@ impl<'a, A: Clone> TreeLike for Partition<'a, A> {
...
+        }
+    }
+}
+
+impl<T: TreeLike> BinaryTree<T> {
+    /// Convert a tree into a binary tree.
+    ///
+    /// ## Panics
+    ///
+    /// `<T as TreeLike>::as_node` returns `Tree::Nary` with an empty array.
```
apoelstra, line 234: In e2e0fd2a0f674e396da6c0e6f5045d86bc362a76:

How can this panic ever happen? It seems like your constructors all prevent this from happening.

apoelstra: Same with a few other `# Panics` sections below.

apoelstra (approved):
ACK 1781c1459745ebd5ba79e3188a7b03544ea4f4cd

src/array.rs
uncomputable: Unfortunately this can happen. `BinaryTree` ultimately wraps an existing `T: TreeLike`. When `tree.as_node()` returns `Tree::Nary(array)` with `array.len() = 0` then there will be a panic in `BinaryTreeInner::array`.

`Tree::Nary` is supposed to be used for nodes with more than two children, but there is nothing that stops implementors from ignoring that convention.

apoelstra: Gotcha, thanks!

PR #18, opened 2024-04-08
move `WitFileData` definition inside the function it is used

apoelstra:
The existing code here is very weirdly structured -- we define a public struct `WitFileData` at the root, then have an `impl` block on it which is *inside* a function (and in the middle of code). The struct in question is only used within the function.

This looks like old code from Sanket and it may be we want to remove it entirely. But for now, having the `pub struct` and `impl` blocks separated causes a new warning in nightly rustc. So pull the struct definition inside the function, remove the `pub`, and rearrange the code.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,11 +34,6 @@ use crate::{
 #[grammar = "minimal.pest"]
 pub struct IdentParser;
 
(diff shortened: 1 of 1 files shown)


uncomputable (approved):
ACK eee867cd13dcbdb279e0173a26d13d906df4fe12

PR #23, opened 2024-04-16
Update Scope

uncomputable:
Enable the scope to track environment values of arbitrary shape (in particular without an initial unit value). This makes it possible to create scopes for functions that expect a specific input type (the function's environment). Compute Simplicity expressions that translate between Simplicity values. The translation expression is compressed.

This is the first step to keep the Simplicity target program and its memory footprint small.

### DIFF
--- a/src/scope.rs
+++ b/src/scope.rs
@@ -4,12 +4,53 @@ use crate::{named::ProgExt, ProgNode};
 
 /// Tracker of variable bindings.
 ///
-/// Internally there is a stack of scopes.
-/// A new scope is pushed for each (nested) block expression.
+/// Each Simfony expression expects an _input value_.
+/// A Simfony expression is translated into a Simplicity expression
+/// that similarly expects an _input value_.
+/// Simfony variable names are translated into Simplicity expressions
+/// that extract the seeked value from the input value of the encompassing expression.
 ///
-/// Bindings from higher scopes (in the stack) overwrite bindings from lower scopes.
+/// Each (nested) block expression introduces a new scope.
+/// Bindings from inner scopes overwrite bindings from outer scopes.
+/// Bindings live as long as their scope.
 #[derive(Debug, Clone)]
 pub struct GlobalScope {
+    /// For each scope, the set of assigned variables.
+    ///
+    /// A stack of scopes. Each scope is a stack of patterns.
+    /// New patterns are pushed onto the top _(current, innermost)_ scope.
+    ///
+    /// The stack of scopes corresponds to a _pattern tree_.
+    ///
+    /// The stack `[[p1], [p2, p3]]` corresponds to a nested product of patterns:
+    ///
+    /// ```text
+    ///    .
+    ///   / \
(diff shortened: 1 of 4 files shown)


uncomputable:
@apoelstra Feel free to take a look

src/scope.rs
```diff
@@ -4,13 +4,59 @@ use crate::{named::ProgExt, ProgNode};
...
+    ///
+    /// p3 ↔ v3
+    /// p2 ↔ v2
+    /// p1 ↔ v1
+    /// ```
+    variables: Vec<Pattern>,
+    /// For each scope, the number of patterns from the outer scope.
+    ///
+    /// When a scope is left, the pattern stack is truncated to this number.
+    scopes: Vec<usize>,
```
uncomputable, line 58: I think I want to 257ab7f5311ab39bcf4de24d9416baa525c55db7. The `scopes` member basically turns the `variables` member back into `Vec<Vec<Pattern>>`. The remaining changes of the PR stay intact.

apoelstra:
In 8a3ec03f3a756dcc0d1133e36416ccc7a80195b6:

I would prefer we explicitly write `GlobalScope::new(Pattern::Ignore)` rather than `GlobalScope::default`. I would even suggest dropping the `Default` impl from `GlobalScope` unless there's a strong usecase for it.

apoelstra:
In 473493b409b0ac2d52912fc4caa7902be40b8af3 (final commit):

I wonder if you could use `Pattern::clone` rather than `.clone` and `.cloned` so that it would be clearer what type is being cloned, and easier for users to look it up and see that it's a cheap (Arc) clone.

src/scope.rs
```diff
@@ -174,18 +188,287 @@ impl Pattern {
...
+            TranslationNode::Unit | TranslationNode::Short(_) => Tree::Nullary,
+            TranslationNode::Pair(l, r) => Tree::Binary(l, r),
+        }
+    }
+}
+
+impl TranslationNode {
+    /// Eliminate an unnecessary pair if possible.
+    /// Return the minimized node.
+    fn eliminate_pair(&self) -> Option<Self> {
```
apoelstra, line 259: In 44ccbdc6ca21f1113042aa53f37d84ef2bc6464e:

This function seems like it's the heart of this commit but I'm struggling to understand it. Maybe we should talk on the phone about it.

uncomputable:
Let's focus on #27 for now and come back to this after Bitcoin++

src/scope.rs
```diff
@@ -11,7 +11,6 @@ use crate::{named::ProgExt, ProgNode};
 #[derive(Debug, Clone)]
 pub struct GlobalScope {
     variables: Vec<Vec<Pattern>>,
-    witnesses: Vec<Vec<WitnessName>>,
```
uncomputable, line 14: Witness names are unscoped but the variables that hold them _are_ scoped. `witness("name")` returns a constant value like a jet `jet_version`.

PR #24, opened 2024-04-17
Add Functions

uncomputable:
Add function declarations and function calls. Do some refactoring.

Depends on #23 

We are discussing function values in general at #20


uncomputable:
It's a little ridiculous to have a long chain of dependent PRs, so let's draft this one.

PR #26, opened 2024-04-23
Make WASM compatible

uncomputable:
Use the latest rust-simplicity master which is compatible with WASM. The version on crates.io is not yet compatible.

Depends on https://github.com/BlockstreamResearch/rust-simplicity/pull/214

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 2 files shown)


uncomputable:
This is ready now @apoelstra. Thanks for merging on the rust-simplicity side

apoelstra (approved):
ACK 970b84147b1e2bf9c22d3b75e2866626bb19a3b1

PR #27, opened 2024-04-23
Recoverable errors

uncomputable:
Replace panics with recoverable results. Print error messages in a similar way as the Rust compiler.

```text
  |
1 | let a1: [bool; 1] = 0x00;
  |                     ^^^^ Value does not match the assigned type `[bool; 1]`
```

Once Simfony has a type system, we can stop triggering low-level Simplicity errors. Users should never see Simplicity errors, much like Rust developers should never see Assembly errors.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,7 @@ pub type ProgNode = Arc<named::NamedConstructNode>;
 mod array;
 pub mod compile;
 pub mod dummy_env;
+pub mod error;
 pub mod named;
 pub mod num;
 pub mod parse;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,7 @@ pub extern crate simplicity;
 pub use simplicity::elements;
 
 use crate::{
+    error::{RichError, WithFile},
     named::{NamedCommitNode, NamedExt},
     parse::{PestParse, Program},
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -36,25 +37,26 @@ use crate::{
...
+        .with_file(program.clone())?;
 
     let mut scope = GlobalScope::new();
-    let simplicity_prog = prog.eval(&mut scope);
-    simplicity_prog
+    let simplicity_program = simfony_program.eval(&mut scope);
+    let named_commit = simplicity_program
         .finalize_types_main()
-        .expect("Type check error")
+        .expect("Type check error");
```
uncomputable, line 50: 2d68ffbae2d051ebe47756dcfef67448e1ab23df: This error is impossible to reach because translated programs always end in `unit`, as of now.

apoelstra:
I'm sorry but I'm going to open a PR that refactors a bunch of stuff and then ask you to redo this.

The existing set of unwraps is a real mess and hard to follow which ones are "legitimate" and which ones should actually be errors that get propagated, and the existing PR rearranges them in ways that I don't think actually fixes the situation.

apoelstra:
https://github.com/BlockstreamResearch/simfony/pull/28

PR #28, opened 2024-04-29
Some error improvements

apoelstra:
Mostly, moves `unwrap`s out of hiding into the compiler where they belong. I did not vet that any of the remaining unwrap()s are actually correct. But I was able to eliminate a ton of them that I was sure *were* correct.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -2,6 +2,7 @@
 
 use std::str::FromStr;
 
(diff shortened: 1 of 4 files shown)


uncomputable (approved):
ACK 722e415a1872e77f58480df5cef79c3128d0afa8 Great changes. I will open a follow-up PR where I extend the cases of infallible combinators.

PR #29, opened 2024-04-29
Update docs

### LINKED ISSUE #16: Arrays and Lists undocumented

uncomputable:
Update documentation of Simfony's syntax. Include arrays, lists, fold and loops as experimental syntax. Fixes #16


doc/syntax.md
```diff
@@ -215,45 +187,107 @@ Then `let p: T = a; b` is an expression
...
+Simfony has familiar types like Boolean and unsigned integers in addition to units, sum and products.
 
-If `c` is an expression
+Simfony's type system is weak in the sense that these nominal types are equivalent to a structural type. For instance, `bool` is equivalent to the sum `Either<(), ()>` of two units. `bool` values are also `Either<(), ()>` values and vice versa. `bool` can be used everywhere `Either<(), ()>` can be used.
 
-If `x` is a variable name
+Nominal types practially don't exist in Simfony's type system. Nominal types are macros for structural types. During compilation, nominal types are resolved to structural types.
 
-If `y` is a variable name
+# Nominal Expressions
```
apoelstra, line 212: In 6edbc8f809fbb8f43d8ebd86be535cfb2a2e3590:

I don't know what "nominal expressions" or "nominal values" are. It looks like you're talking about literal expressions. You can [take a look at the Rust docs](https://doc.rust-lang.org/reference/expressions/literal-expr.html) for some inspiration about how to describe these.

doc/syntax.md
```diff
@@ -215,45 +187,107 @@ Then `let p: T = a; b` is an expression
...
+| `u32`          | `(u16, u16)`     | 32-bit unsigned integer      | `0`, `1`, ..., `4294967295`           |
+| `u64`          | `(u32, u32)`     | 64-bit unsigned integer      | `0`, `1`, ..., `18446744073709551615` |
+| `u128`         | `(u64, u64)`     | 128-bit unsigned integer     | strings of 32 hex digits              |
+| `u256`         | `(u128, u128)`   | 256-bit unsigned integer     | strings of 64 hex digits              |
 
-If `b` is an expression
+Simfony has familiar types like Boolean and unsigned integers in addition to units, sum and products.
 
-If `c` is an expression
+Simfony's type system is weak in the sense that these nominal types are equivalent to a structural type. For instance, `bool` is equivalent to the sum `Either<(), ()>` of two units. `bool` values are also `Either<(), ()>` values and vice versa. `bool` can be used everywhere `Either<(), ()>` can be used.
```
apoelstra, line 208: In 6edbc8f809fbb8f43d8ebd86be535cfb2a2e3590:

Lol. Yeah, I think this is okay. The proper way to describe this is that Simfony simply doesn't have nominal types. It just has structural types and you're able to give multiple names to the same types. But because we are aspirationally going to support nominal types, this sort of language ("it has nominal types but the type system is so weak they cannot be observed") is probably fine.

apoelstra:
Overall d25f5334296262ba046c3dc6915589e3ba9320f7 looks good. Definitely an improvement over the old outdated docs. I think the `List` type could maybe be described more clearly by analogy to binary numbers but I can't think of language that would be clear.

I guess at some point we'd like to generalize `forWhile` a bit so you don't need to use 2^2^n but that's separate from this PR.

apoelstra:
Also the last commit d25f5334296262ba046c3dc6915589e3ba9320f7 has trailing whitespace somewhere . You can find it with `git grep ' $'`.

doc/syntax.md
uncomputable: You're right, Simfony doesn't have nominal types yet. Simfony has type macros that are resolved to structural types. The expressions that work on type macros are themselves macros. I updated the documentation.

doc/syntax.md
uncomputable: What we have in Simfony are actually macro expressions and, I guess, "macro values".

apoelstra (approved):
ACK 91633efe33a31b963ab913ee273243e7917da7ca

apoelstra:
Ok, great, 91633efe33a31b963ab913ee273243e7917da7ca looks much better. I think this "macro" language maybe leaks too much of the implementation (like, does the user really care that Simfony expressions are "macroized" into Simplicity? and will this even continue to be true when we add optimization passes?).

In Miniscript we have an explicit design goal that every Miniscript node corresponds exactly to a fragment of Script. But here I guess, we want the *semantics* of Simfony expressions to exactly match those of Simplicity.  But they don't necessarily need to compile down to exactly-specified Simplicity.

uncomputable:
I think we will drop the macro language once we drop macro types. I used the word "macro" because a macro and its resolved type/value/expression are interchangeable, which is potentially confusing.

Once we reorder variable declarations and drop variables as soon as we can, expressions will no longer have exactly one translation to Simplicity. An expression in one part of the program can influence the translation of another part of the program.

PR #32, opened 2024-05-03
Pairs of {O,I,H}

uncomputable:
(Nested) pairs of take, drop and iden always type-check. Add a builder for these pairs.


apoelstra (approved):
ACK 7970208a9abe5aa5d51b59aa66632165bae54ca9 nice, I buy this argument

PR #37, opened 2024-05-14
Infallible assertions

uncomputable:
Introduce infallible assertions and use them in the compiler.

It turns out, assertions _can_ fail if the source type of the child is a unit or sum type. No upstream fix is needed.


apoelstra (approved):
ACK 780364dd9c085b29b75d73e9f764eeabde228728

PR #38, opened 2024-05-23
Generalize Scope

uncomputable:
Generalize the scope to return values of multiple variables inside a pattern. We will use these values as function call arguments. The seeked value is returned by a Simplicity expression that is compressed.

Add supporting structs and documentation.

Supersedes #23

### DIFF
--- a/src/scope.rs
+++ b/src/scope.rs
@@ -412,3 +412,34 @@ impl Pattern {
             .map(PairBuilder::get)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn translate_pattern() {
+        let a = BasePattern::Identifier(Identifier::from_str_unchecked("a"));
+        let b = BasePattern::Identifier(Identifier::from_str_unchecked("b"));
+        let c = BasePattern::Identifier(Identifier::from_str_unchecked("c"));
+        let env = BasePattern::product(BasePattern::product(a.clone(), b.clone()), c.clone());
+
+        let target_expr = [
+            (a.clone(), "OOH"),
+            (b.clone(), "OIH"),
+            (c.clone(), "IH"),
+            (BasePattern::product(a.clone(), b.clone()), "OH"),
+            (BasePattern::product(a.clone(), c.clone()), "OOH & IH"),
+            (BasePattern::product(b.clone(), a.clone()), "take (IH & OH)"),
+            (BasePattern::product(b.clone(), c.clone()), "OIH & IH"),
+            (BasePattern::product(c.clone(), a.clone()), "IH & OOH"),
+            (BasePattern::product(c.clone(), b.clone()), "IH & OIH"),
+            (env.clone(), "iden"),
+        ];
+
+        for (target, expected_expr) in target_expr {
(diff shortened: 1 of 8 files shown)


src/scope.rs
```diff
@@ -156,6 +157,42 @@ impl BasePattern {
...
+    /// The expression is a sequence of `take` and `drop` followed by `iden`.
+    fn get(mut self: &Self, identifier: &Identifier) -> Option<SelectorBuilder<ProgNode>> {
+        let mut selector = SelectorBuilder::default();
+        loop {
+            // Termination: self becomes strictly smaller in each iteration
+            match self {
+                BasePattern::Identifier(self_id) if self_id == identifier => return Some(selector),
+                BasePattern::Identifier(_) | BasePattern::Ignore => return None,
+                BasePattern::Product(self_left, self_right) => {
+                    if self_left.contains(identifier) {
```
apoelstra, line 195: In fb212840cc393d16ed2961721eb93650bec3c8d8:

I agree that the old `DirectedTree` code was overly general and hard to read. But this code has worse asymptotics because every call to `contains` iterates through the entire (sub)tree, so the algorithm winds up being O(n^2). It would be better if `contains` would record the path that it takes so that we don't need to call it more than once.

I would suggest using `verbose_pre_order_iter` in `contains` , and returning a vector of left/right decisions which you can pop every time you hit a `BasePattern::Product` in this loop. I'd also make `contains` non-public or rename it to reflect the richer return value.

Fine to do in a followup PR since it's not really related to this one.

src/scope.rs
```diff
@@ -225,6 +225,153 @@ impl BasePattern {
...
+        self.pre_order_iter().flat_map(BasePattern::as_identifier)
+    }
+
+    /// Check if all `identifiers` are contained inside the pattern.
+    pub fn contains_all<'a, I>(&self, mut identifiers: I) -> bool
+    where
+        I: Iterator<Item = &'a Identifier>,
+    {
+        identifiers.all(|id| self.contains(id))
+    }
```
apoelstra, line 252: In 8477e265153d4864f228caa4c819e77f91e678bb:

This implementation of `contains_all` also has bad asymptotics. It shouldn't call `contains` repeatedly. Instead you should construct a hashset of all the identifiers in `self` and then check that every item yielded from the iterator lands in the hashset.

apoelstra (approved):
ACK 4e83e846a652f512efa829211f1d0bfff06ea286 whew. I did not follow the logic in super-close detail but I think I understand what this PR is doing. The docs/pictures in the final commit help. Looks good except for the performance comments, which can be addressed later.

src/scope.rs
uncomputable: `contains` iterates over the tree in pre-order short-citcuits on the first match. This is exactly what `DirectedTree` does.

The problem is `contains_all` which repeatedly calls `contains`, making it O(n^2). Let's address the performance in a follow-up PR. Thanks for your suggestion to use a hashset.

uncomputable: Oh, I see that `get` also calls `contains`. This is indeed slower than `DirectedTree`. Again, thanks for your suggested fix. We might need different `contains` methods for `contains_all` and for `get`.

PR #41, opened 2024-05-28
Explicit types

uncomputable:
Require explicit types in each let statement.

Simfony needs to know the type of each expression for proper compilation and error reporting. The simplest approach is [manifest typing](https://en.wikipedia.org/wiki/Manifest_typing) where the programmer provides the type of each expression. The Simfony compiler can do equality checks and simple pattern matching to check the types.


apoelstra (approved):
ACK ecd11b8f8ece5b4f856e00c1dfea7d8878d5364f

PR #42, opened 2024-06-06
Type module

uncomputable:
Move types into their separate module. Rename `Type` to `ResolvedType` to better reflect its function in face of future aliased types. Introduce `StructuralType` to represent Simplicity types in the code base. Implement `From` and `TryFrom` to inter-type conversions. Introduce `TypeInner` to handle type primitives of both resolved and future aliased types.


apoelstra (approved):
ACK 20a204b47e24ed30794524545457f0a0ae6bb299

PR #43, opened 2024-06-06
Update byte strings

uncomputable:
Make byte strings compatible with Rust by disallowing nibbles and strings longer than 16 bytes. Display erratic length for easier debugging. Allow underscores for visual separation inside the string.


src/error.rs
```diff
@@ -169,7 +169,7 @@ impl fmt::Display for Error {
             ),
             Error::HexStringPow2 => write!(
                 f,
-                "Length of hex string must be a power of two: 1, 2, 4, 8, 16, ..."
+                "Length of hex string must be a power of two greater one: 2, 4, 8, 16, 32, ..."
```
apoelstra: In 1163ab7148f4c9712fc35cd34d9b6cb07f4cc83d:

Should be `greater than one`.

src/error.rs
```diff
@@ -157,19 +157,19 @@ impl fmt::Display for Error {
         match self {
             Error::ArraySizeZero => write!(
                 f,
-                "Array size must be positive: 1, 2, 3, 4, 5, ..."
+                "Array size must be positive: 1, 2, 3, 4, 5, ... Got 0"
```
apoelstra, line 160: In 23611c2b3757ca5ac0188b72ddac700b2b87e9f7:

Since I am nitting grammer, `got` should be lowercase in all these.

apoelstra:
Done reviewing 490e388e1eea1882239b52eb753407b008ace1a6. I guess this is ok. Rust's awful support for hexadecimal literals is one of its worst syntactic problems, and possibly the only one that has continually bothered me for over a decade without ever getting better or my getting acclimated to it.

So I guess, if this is the only compromise we make for "Rust compatibility" that's okay.... but it's a really big compromise and it makes me want to drop that goal, so we should be mindful if we find ourselves making other bad syntactic choices like this.

ACK other than grammar nits.

uncomputable:
I expect that Simfony will never be 100% syntactic Rust because of missing generics or something else that Simfony doesn't support. There will be a mechanical translation of Simfony source code to Rust source code, where we could also convert longer hex literals into shorter, Rust-compatible ones.

uncomputable:
I was planning to remove `u256` in favor of `[u8; u32]` or more specialized types like `PublicKey`, but I just realized that this will make Simplicity's 256-bit arithmetic jets really hard to use. It is really annoying to manipulate 256-bit values in Rust. So, I chose to keep `u256` and its hex literals.

apoelstra (approved):
ACK 31d8fa0b7f31204c8e559b77bb906f35dfe603c9

PR #44, opened 2024-06-08
Type alias

uncomputable:
Add type aliases.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -7,6 +7,7 @@ use simplicity::types::{CompleteBound, Final};
 
 use crate::array::{BTreeSlice, Partition};
 use crate::num::NonZeroPow2Usize;
+use crate::parse::Identifier;
 
 /// Primitives of the Simfony type system, excluding type aliases.
 #[derive(Debug, PartialEq, Eq, Hash, Clone)]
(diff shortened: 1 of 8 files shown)


src/types.rs
```diff
@@ -258,6 +259,149 @@ impl fmt::Display for ResolvedType {
...
+        Self(AliasedInner::Alias(identifier))
+    }
+
+    /// Resolve all aliases in the type based on the given map of `aliases` to types.
+    pub fn resolve<F>(&self, mut get_alias: F) -> Result<ResolvedType, Identifier>
+    where
+        F: FnMut(&Identifier) -> Option<ResolvedType>,
+        // The borrow checker doesn't let me return Option<ResolvedType>
+        // F: FnMut(&Identifier) -> Option<&ResolvedType>,
+        // note: due to current limitations in the borrow checker, this implies a `'static` lifetime
```
apoelstra, line 289: In 43db8c9aa2b4d85d70706d1eee97e18a9618cc53:

FYI if you add a `'s` lifetime parameter to `resolve`, change `&self` to `&'s self`, then change `&Identifier` to `&'s Identifier` and `Option<ResolvedType>` to `Option<&'s ResolvedType>`, things work.

But this doesn't really gain you anything. It lets you drop the `.cloned()` on the closure you pass to `resolve` but then you need to add a `.cloned()` within `resolve`. There's no way to return a reference from `resolve` itself since sometimes you construct new `ResolvedType`s and return it.

I would suggest just dropping this comment because it invites people to try to mess around with it, and the problem isn't actually with the borrowchecker.

apoelstra (approved):
ACK 7463a4f74e701243401bf8a8478660c45a99a7c1 other than the comment thing; and I observe that this means that you need to define before they are used...which is fine, but feels a little 1970sish

uncomputable:
Rebased and removed the misleading comment.

The compiler processes the Simfony source code in one sequence. This means that types have to be defined before they are used. The same will hold for functions once I add them. The developer experience is not ideal, but it should be good enough for version 1.

apoelstra (approved):
ACK 7b2e18b6e524e9e0a5d47192938c6671c9402adc

PR #45, opened 2024-06-09
Decimal literals

uncomputable:
Add support for decimal literals to `u128` and `u256`. This means that all unsigned integer types now support bit, byte and decimal literals. Introduce the `U256` struct to handle 256-bit unsigned integers.


apoelstra (approved):
ACK cabecec337529ca781a6aea9748a9aafbd81b83f LGTM. I guess we still need `Display`..but probably also `Add`, `Sub`, `Mul`, etc?

uncomputable:
For now, 256-bit arithmetic is not required, because there is no pure Simfony runtime. Simfony programs are compiled to Simplicity and run inside the Simplicity runtime.

PR #46, opened 2024-06-14
Implement fmt::Display for U256

uncomputable:


apoelstra (approved):
ACK 8650b29e0f27cbd40d24b63d20fe4a6296fa8b0e eventually we should make this allocation-free but this is fine for now

uncomputable:
Do you mean we should use `[u8; 78]` instead of `Vec::with_capacity(78)`?

apoelstra:
@uncomputable oh, lol, yeah, that's a great idea.

I had meant that we should change the whole algorithm somehow to compute the digits in the correct order. But just replacing the vec with an array is way simpler and probably more efficient.

PR #47, opened 2024-06-17
Value cleanup

uncomputable:
Cleanup of the value / type infrastructure in preparation of adding proper Simfony values to the code base.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -373,13 +373,22 @@ impl fmt::Display for UnsignedDecimal {
 /// Bit string whose length is a power of two.
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub enum Bits {
-    /// Least significant bit of byte
+    /// Least significant bit of byte.
+    ///
+    /// The bit is saved as a [`u8`] value that is always less than or equal to 1.
     U1(u8),
-    /// Two least significant bits of byte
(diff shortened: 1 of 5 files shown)


src/parse.rs
```diff
@@ -373,13 +373,22 @@ impl fmt::Display for UnsignedDecimal {
 /// Bit string whose length is a power of two.
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub enum Bits {
-    /// Least significant bit of byte
+    /// Least significant bit of byte.
+    ///
+    /// The bit is saved as a [`u8`] value that is always less equal 1.
```
apoelstra, line 378: In 980a894f9c05336579c91ab02ada69c4dc74fbb0

`less equal` should be `less than or equal to` or `<=` everywhere.

apoelstra (approved):
ACK 80c906f1325875b776da91656fac7440312c6f3a other than the grammar nit

src/parse.rs
apoelstra: Yeah, in English there is no real short way to say this. Often you will hear "no greater than" which is shorter than "less than or equal to" but still a bit of a mouthful.

apoelstra (approved):
ACK 5c3d46c2ea3bfe7e3ace5c3a3a0d6763e7d191f1

PR #48, opened 2024-06-17
Add values

uncomputable:
Add representation of values in correspondence to the newly added type module:

- `UIntValue` ↔ `UIntType`
- `Value` ↔ `ResolvedType`
- `StructuralValue` ↔ `StructuralType`
- `ValueConstructible` ↔ `TypeConstructible`

Depends on #47

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -1,0 +1,193 @@
+use std::fmt;
+
+use crate::error::Error;
+use crate::num::U256;
+use crate::parse::{Bits, Bytes, UnsignedDecimal};
+use crate::types::UIntType;
+
+/// Unsigned integer value.
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub enum UIntValue {
+    /// 1-bit unsigned integer.
+    U1(u8),
+    /// 2-bit unsigned integer.
+    U2(u8),
+    /// 4-bit unsigned integer.
+    U4(u8),
+    /// 8-bit unsigned integer.
+    U8(u8),
+    /// 16-bit unsigned integer.
+    U16(u16),
+    /// 32-bit unsigned integer.
+    U32(u32),
+    /// 64-bit unsigned integer.
+    U64(u64),
+    /// 128-bit unsigned integer.
+    U128(u128),
+    /// 256-bit unsigned integer.
+    U256(U256),
+}
+
+impl fmt::Display for UIntValue {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            UIntValue::U1(n) => <u8 as fmt::Display>::fmt(n, f),
+            UIntValue::U2(n) => <u8 as fmt::Display>::fmt(n, f),
+            UIntValue::U4(n) => <u8 as fmt::Display>::fmt(n, f),
+            UIntValue::U8(n) => <u8 as fmt::Display>::fmt(n, f),
+            UIntValue::U16(n) => <u16 as fmt::Display>::fmt(n, f),
+            UIntValue::U32(n) => <u32 as fmt::Display>::fmt(n, f),
+            UIntValue::U64(n) => <u64 as fmt::Display>::fmt(n, f),
+            UIntValue::U128(n) => <u128 as fmt::Display>::fmt(n, f),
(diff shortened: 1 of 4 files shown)


src/value.rs
```diff
@@ -0,0 +1,188 @@
...
+            32 => Ok(Self::U256(U256::from_byte_array(value.try_into().unwrap()))),
+            _ => Err("Too many bytes"),
+        }
+    }
+}
+
+impl From<Bits> for UIntValue {
+    fn from(value: Bits) -> Self {
+        match value {
+            Bits::U1(byte) => Self::u1(byte).expect("Always <= 1"),
```
apoelstra, line 181: In a786308a9bb84f7dc2e0b82dd3aec23d72bd9f81:

FWIW this isn't actually guaranteed because `Bits` is a public enum and users can construct it wrong if they want.

src/value.rs
```diff
@@ -186,3 +189,267 @@ impl From<Bytes> for UIntValue {
...
+    /// Create an array from the given `elements`.
+    ///
+    /// ## Nonemptiness
+    ///
+    /// There must be at least one element.
+    ///
+    /// ## Errors
+    ///
+    /// Returns [`None`] if there are no elements.
+    // TODO: Enforce the nonemptiness of the iterator through the type system?
```
apoelstra, line 218: In 892062c29dcc3787f20539b31a7bc153c29ae507
 
I'd just drop this TODO (and the one below). I've tried to enforce conditions like this with the type system and it just winds up being super unergonomic in Rust.

src/value.rs
```diff
@@ -186,3 +189,267 @@ impl From<Bytes> for UIntValue {
...
+    /// Option value.
+    Option(Option<Arc<Self>>),
+    /// Boolean value.
+    Boolean(bool),
+    /// Unsigned integer.
+    UInt(UIntValue),
+    /// Nonempty array of values.
+    Array(Arc<[Self]>),
+    /// Bounded list of values.
+    List(Arc<[Self]>, NonZeroPow2Usize),
```
apoelstra, line 255: In 892062c29dcc3787f20539b31a7bc153c29ae507:

This `eum Value` allows the user to construct invalid values for the `List` and `Array` variants. That may be okay (or at least, too annoying to prevent) but we should be aware of it. The "correct" way to handle this is to move the enum into an `enum ValueInner` and have the public interface go through an opaque newtype around that with constructors that enforce validity.

src/value.rs
```diff
@@ -427,6 +427,50 @@ impl Value {
...
 
+/// Typed Simfony value.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct TypedValue {
+    value: Value,
+    ty: ResolvedType,
+}
+
+impl TypedValue {
+    /// Create a `value` of the given `ty`pe.
```
apoelstra, line 440: In 03b24212d95fdd60b3e23cbdd510911bcabe2b2d:

typo `type`

apoelstra:
Done reviewing 4572cdc1879a949c78dd7a2f9d80c72c09d34237

src/value.rs
uncomputable: Mmh. We might want to force users through a constructor that checks the invariants by making the enum private...

src/value.rs
uncomputable: I see you don't like my attempt at highlighting the parameter `ty` which cannot be called `type` because that would be a keyword :) I admit that "`ty`pe" looks weird and should be just "type".

apoelstra: Oh, lol :) yeah, I see what you were trying to do. But sadly it's not obvious to a reader. I think you should just use "type" without any backticks.

src/value.rs
uncomputable: In theory any property of a value can be a type, but Rust's type system has its limits.

src/value.rs
uncomputable: Preventing invalid array or list values would be annoying because we still want to do pattern matching. For now, I will add a comment that this is an issue.

We could add methods like `as_array()` and `as_list()` that inspect the value instead of doing pattern matching. There are not so many value variants.

uncomputable:
Rebased and made sure that bit and byte strings uphold the claimed invariants.

apoelstra (approved):
ACK ae863586b882f29fa76f34c547284a7ddb1ca513

PR #49, opened 2024-06-20
Parse cleanup

uncomputable:
Cleanup in the parse module.

PR #50, opened 2024-06-21
Add tuples

uncomputable:
Add tuple types and values. We need tuples to represent function arguments, which have heterogeneous types. Up to this point, we abused arrays for this purpose, because tuple and arrays have the same Simplicity structure. When we give Simfony proper types, this hack will no longer work.

Tuples subsume unit and product: Unit is the empty tuple and a product is a 2-tuple. For now, keep the grammar as before, which means only 0-tuples and 2-tuples are allowed.

Depends on #49

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -555,3 +555,31 @@ impl StructuralType {
         simplicity::types::Type::from(self.0.clone())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn display_type() {
+        let unit = ResolvedType::unit();
+        assert_eq!("()", &unit.to_string());
+        let singleton = ResolvedType::tuple([ResolvedType::from(UIntType::U1)]);
+        assert_eq!("(u1,)", &singleton.to_string());
+        let pair = ResolvedType::tuple([
+            ResolvedType::from(UIntType::U1),
+            ResolvedType::from(UIntType::U8),
+        ]);
+        assert_eq!("(u1, u8)", &pair.to_string());
+        let triple = ResolvedType::tuple([
+            ResolvedType::from(UIntType::U1),
+            ResolvedType::from(UIntType::U8),
+            ResolvedType::from(UIntType::U16),
+        ]);
+        assert_eq!("(u1, u8, u16)", &triple.to_string());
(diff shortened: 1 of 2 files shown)


uncomputable:
Rebased

src/types.rs
```diff
@@ -63,6 +65,25 @@ impl<A> TypeInner<A> {
...
+            TypeInner::Tuple(elements) => match n_children_yielded {
+                0 => {
+                    f.write_str("(")?;
+                    if 0 == elements.len() {
+                        f.write_str(")")?;
+                    }
+                    Ok(())
+                }
+                n if n == elements.len() => {
+                    if n == 1 {
```
apoelstra, line 64: In a01cfa130f0bab5d370bf9781bc133ca05d10133:

I think this should be `if n != 0`.

src/types.rs
```diff
@@ -555,3 +555,25 @@ impl StructuralType {
...
+    fn display_type() {
+        let unit = ResolvedType::unit();
+        assert_eq!("()", &unit.to_string());
+        let singleton = ResolvedType::tuple([ResolvedType::from(UIntType::U1)]);
+        assert_eq!("(u1,)", &singleton.to_string());
+        let tuple = ResolvedType::tuple([
+            ResolvedType::from(UIntType::U1),
+            ResolvedType::from(UIntType::U8),
+        ]);
+        assert_eq!("(u1, u8)", &tuple.to_string());
```
apoelstra, line 573: In 7b7bac7afd0c63871f8bf4c13b2866894015995a:

Should have a test with tuples with more than 2 items.

src/types.rs
uncomputable: The case `n == 1` is for printing the trailing comma in `(1,)` as opposed to printing `(1)`. Normal (nontrailing) commas as in `(1, 2)` are printed in the match arm `n` where `n != elements.len()`.

src/types.rs
uncomputable: I added a test for triple types and values. Hopefully this clarifies that the Display code is correct.

apoelstra (approved):
ACK 32cdc61b0651d0b697f40227f0f8e7148220d8be

PR #53, opened 2024-07-03
AST cleanup

uncomputable:
Clean up the code in preparation of the abstract syntax tree, which we will add next. A lot of the diff is moved code, which will shrink the diff in the PR that adds the actual AST.


apoelstra (approved):
ACK 710a4c393ff0a28fa4af0eda263fe543a6e76978

PR #54, opened 2024-07-04
Jets

uncomputable:
Assign a Simfony type signature to each Simplicity jet. The Simfony signature can be arbitrarily different from the Simplicity signature, as long as the compiler can convert between the two. In this commit, I chose signatures that are structurally equivalent, so no conversion is necessary. The equivalence is asserted in a unit test.

## Nonstandard tuples

Simfony converts arrays and tuples into binary trees and always prefers the left branch. For instance `(a, b, c)` becomes `((a, b), c)`. This is what I call a "standard tuple". Simplicity sometimes prefers the left branch, sometimes the right branch. `(a, b, c)` can become `(a, (b, c))`, which I call a "nonstandard tuple".

For example, there is the Simplicity jet `maj1: 2 × (2 x 2)` which becomes `fn maj1(a: u1, b: (u1, u1))` in Simfony. `fn maj1(a: u1, b: u2, c: u3)` would be more natural, but it requires a conversion.

I think the preference of the left or of the right branch comes from the way how the jets are nested as fully expanded Simplicity expressions. Always preferring the left branch might lead to larger expressions and other problems. In this case, it seems best to update the Simfony signature of the affected jets and to teach the compiler to insert a converter expression between the arguments and the actual jet (in a future PR).

## Type constructors

Methods from `<AliasedType as TypeConstructible>` cannot be imported. These methods also require `AliasedType` as a parameter, so we cannot write `AliasedType::option(U1)` but we have to write `AliasedType::option(U1.into())`. `jets.rs` is prefixed with methods that take `A: Into<AliasedType>` as paramater, so we can write `option(U1)`. The prefix `AliasedType::` is also omitted. I am not sure if we should update `TypeConstructible` to include `A: Into<AliasedType>`.

## Allocation vs const

I would like the jet source and target types to be compile time constants. However, this is currently not possible because types consist of `Arc`s. Without `lazy_static`, we have to generate the types from new every time and return a vector of them. Not ideal, but maybe acceptable.

### DIFF
--- a/example_progs/ctv.simf
+++ b/example_progs/ctv.simf
@@ -6,8 +6,6 @@
  * that they want to commit.
  */
 
-type Ctx8 = (List<u8, 64>, (u64, u256));
-
 let ctx: Ctx8 = jet_sha_256_ctx_8_init();
 let ctx: Ctx8 = jet_sha_256_ctx_8_add_4(ctx, jet_version());
 let ctx: Ctx8 = jet_sha_256_ctx_8_add_4(ctx, jet_lock_time());
--- a/example_progs/sighash_all.simf
+++ b/example_progs/sighash_all.simf
@@ -9,8 +9,6 @@
  * the value that was returned from the dedicated jet.
  */
 
-type Ctx8 = (List<u8, 64>, (u64, u256));
-
 // jet_sig_all_hash
 let ctx: Ctx8 = jet_sha_256_ctx_8_init();
 let ctx: Ctx8 = jet_sha_256_ctx_8_add_32(ctx, jet_genesis_block_hash());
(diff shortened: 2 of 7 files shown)


roconnor-blockstream:
In the Haskell term language, it is (kinda) right associated:

https://github.com/BlockstreamResearch/simplicity/blob/master/Haskell/Core/Simplicity/Term/Core.hs#L77-L82

i.e. `s &&& t &&& u : A ⊢ B × (C × D)` where `(&&&)` is the infix version of `pair`.

uncomputable:
Ok, I chose left-associativity for Simfony without giving it much thought, simply because left comes before right. We can easily change Simfony to right-associativity.

apoelstra (approved):
ACK 4e5657533115cfc28a59d416f13b32f363129a34 Did not carefully check that every type is correct. Upstream we should add a method to iterate over all jets and to "index" into the list of jets. I needed this for the fuzzer as well

PR #55, opened 2024-07-04
Codegen

uncomputable:
Depends on #54 

Add documentation for each jet. The documentation is currently not needed in Simfony itself, so it lives in a subcrate that generates Rust code for https://github.com/uncomputable/simfony-as-rust, a compatibility layer to work on Simfony inside the Rust ecosystem. We can use rustdoc to generate [documentation for each jet](https://docs.rs/simfony-as-rust/latest/simfony_as_rust/).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,6 +93,13 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
+name = "codegen"
+version = "0.1.0"
+dependencies = [
+ "simfony",
+]
+
+[[package]]
 name = "cpufeatures"
 version = "0.2.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


uncomputable:
Rebased

apoelstra:
concept ACK but (a) where did this text come from? Can you provide instructions or pointers on how to reproduce it? and (b) can you run `cargo +nightly fmt` over it?

Unrelatedly, one benefit of the local-CI overhaul I did 2 months ago is that it now detects new crates and runs all the tests on them, including generating a feature matrix, without any intervention from me :).

uncomputable:
This text cannot be auto generated because it was written by hand. Sources are the tech report, especially for the Elements jets; the [Haskell documentation](https://github.com/BlockstreamResearch/simplicity/blob/master/Haskell/Core/Simplicity/Programs/LibSecp256k1.hs), especially for the libsecp jets; and conversations with @roconnor-blockstream. Truth is, there is no single source of documentation for all jets. I made it here.

uncomputable:
I ran the code formatter.

apoelstra (approved):
ACK eb63a79f17f706967b9276398ea99fe71873165a

PR #60, opened 2024-07-09
Examples: Write Ctx8

uncomputable:
Compress the example program text by using the `Ctx8` type alias.


apoelstra (approved):
ACK 21cd2427fe34ed78c0494d7da9b550b33ae9f60c

PR #64, opened 2024-07-17
Add functions

uncomputable:
Depends on #63 Better version of #24 

Move the existing parts of the program into the `main` function (except type aliases). Add the possibilities to define more functions outside the `main` function. Functions can be called _after_ they are defined. This feels like 1970, but it keeps the compiler simple.

Witness expressions are only allowed inside the `main` function to prevent indirect witness reuse via multiple function calls.


uncomputable:
Rebased and ready for review

apoelstra (approved):
ACK 5621179e7a074242d2e36bd6baaeba26a4ba3b38

PR #65, opened 2024-07-18
Add type casts

uncomputable:
Depends on #64 

Add type casts based on structural type equality.


uncomputable:
Rebased and ready for review

uncomputable:
Ran `rustfmt`. Sorry for forgetting that.

PR #73, opened 2024-08-11
lockfile: reduce dependency versions to ones that MSRV can handle

apoelstra:
I recently updated my local CI to run `cargo test --doc` in addition to the existing tests. Inexplicably, this is causing it to try to bulid more dependencies than it previously was, and it turns out that some dependencies don't build on MSRV.

We have committed a lockfile in this repo, which is inadvisable for a library, but since we did it, my local CI setup tries to use it. So update it to something that works.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.0.5"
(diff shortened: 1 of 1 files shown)


uncomputable (approved):
ACK 0ed932e98bbaf61d1fe1176296c736b020a0fe86

Issue #72, opened 2024-08-11
Fuzz compiler


uncomputable:
We should check if the compiler can handle malicious inputs and present an error instead of crashing.

apoelstra:
```diff
diff --git a/src/compile.rs b/src/compile.rs
index c035918..6671fc9 100644
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -434,3 +434,11 @@ impl Match {
         ProgNode::comp(&input, &output).with_span(self)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn fuzz_regression_1() {
+        crate::compile("typef=f").unwrap_err();
+    }
+}
```

This crashes with an assertion failure
```

---- compile::tests::fuzz_regression_1 stdout ----
thread 'compile::tests::fuzz_regression_1' panicked at src/parse.rs:764:9:
assertion failed: matches!(pair.as_rule(), Self::RULE)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I found this using the following fuzz target

```rust
#![no_main]

use libfuzzer_sys::fuzz_target;

fuzz_target!(|data: &[u8]| {
    let _ = core::str::from_utf8(data).map(simfony::compile);
});
```

Which I set up using the instructions for [cargo fuzz](https://github.com/rust-fuzz/cargo-fuzz).

PR #74, opened 2024-08-13
Fuzz fixes

uncomputable:
I set up the fuzzing infrastructure as described in #72 and looked for crashes. I fixed the bug that Andrew found in #72. There were no more crashes, but it turns out that parsing can be really show when opening `(`, `[`, `{` are spammed and not closed. This seems to be an issue with the PEST parser itself. I will look into fixing this, but for now I will just note that parsing can be slow. Unfortunately, this also means that fuzzing is slow. I found no way to skip slow units during fuzzing. I limited the input size to around 50 bytes, but even this produces slow units.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,3 @@
-workspace = { members = ["codegen"] }
 [package]
 name = "simfony"
 version = "0.1.0"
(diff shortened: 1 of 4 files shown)


apoelstra:
In a520762e71ccf3fb949aed622e1fb3b0ae142f22:

We should not be requiring users use a literal space after keywords. In Rust you can use whatever whitespace you want, and in the case of `let` bindings, even no whitespace (e.g. in `let(x,y) = (1, 2)`.

uncomputable:
I updated the code to no longer enforce whitespaces.

apoelstra (approved):
ACK f21b49523b50cfc4211734bcceb245c1768129ab successfully ran local tests

PR #75, opened 2024-08-13
Add for-while loops

uncomputable:
Implement for-while loops based on [the existing Haskell code](https://github.com/BlockstreamResearch/simplicity/blob/121b2951d8cca3d73dccd6fcbe721f2195025067/Haskell/Core/Simplicity/Programs/Word.hs#L354-L363). _(Function signatures are adapted for right-associativity for a better fit with the rest of Simfony.)_ Implement for-while iteratively such that the exact same Simplicity expressions are returned as for the recursive implementation. Updatre the number types to support this implementation.


uncomputable:
Updated one "Safety" comment.

src/num.rs
```diff
@@ -64,6 +64,15 @@ impl NonZeroPow2Usize {
         }
     }
 
+    /// Create a power of two with nonzero exponent.
+    ///
+    /// ## Safety
+    ///
+    /// The value must be a power of two with nonzero exponent.
+    pub const unsafe fn new_unchecked(n: usize) -> Self {
```
apoelstra, line 72: In 97c0ebc214c8e55a1c43939b957259e8a67dde3f:

Why is this unsafe?

uncomputable: I copied the pattern from [NonZeroUsize](https://doc.rust-lang.org/std/num/struct.NonZero.html#tymethod.new_unchecked). I understood `unsafe` as a marker for a precondition that must be upheld for the code to work. In the case of Simfony's number types, the number must be a power of two and so on.

If `unsafe` should only be used for number types of the stdlib, which enable compiler optimizations, then I can remove `unsafe` from Simfony's number types. I can use a `debug_assert!` instead.

apoelstra: > I understood `unsafe` as a marker for a precondition that must be upheld for the code to work

No, it means a marker for a precondition that must be upheld for the code to be sound.

`NonZeroUsize` is actually unsound if you put a 0 into there. It will cause undefined behavior.

apoelstra: So yes, I would recommend removing the `unsafe` marker. (Alternately you can change the internal type to use `NonZeroUsize` and then you can reuse the compiler optimization...but given that you can only represent like, 64 different values, it seems silly to introduce unsafe code just to tell the compiler that one is available. It would be better to store the exponent in a `u8` or a `NonZeroU8` if you want to try to optimize storage.)

apoelstra:
79d4c16e5e0a8b65e6002d3532d7e9ddf2b69c8a looks good other than the `unsafe` thing.

I also wish that the user didn't have to explicitly indicate the counter type in the for-loop. It seems like we should be able to infer it from the type of the function passed?

uncomputable:
Rebased, removed `unsafe` and removed the generic bit width parameter from `for_while`.

uncomputable:
Updated a commit description

apoelstra (approved):
ACK 2ae2cae63bf70a4934688d9e5dec117c25a02702 successfully ran local tests

PR #76, opened 2024-08-20
Update jet docs

uncomputable:
Fix some issues that came up while using the docs.


apoelstra (approved):
ACK 409b982681f527789745c660100247d3faea9e4a successfully ran local tests

PR #77, opened 2024-08-20
Update parse tree

uncomputable:
Prepare the parse tree for fuzzing, which will come in a follow-up PR.

Move string types into a separate module. Implement `PartialEq + Eq + std::hash::Hash` for Simfony tree structures such that `Span` is ignored. Implement `Display` for the parse tree in an iterative fashion. Fix bugs that were discovered while fuzzing.

Lots of moves lines of code and mechanical changes.


apoelstra:
In e34e58ee076ae8efcdc5eed45d48e9a9e422f0af:

nit: the commit message says you implement `Display` with a post-order iterator, but of course you actually use a pre-order iterator.

apoelstra (approved):
ACK e3370392568dc9974e1d38f0c2c1444938d4f564 successfully ran local tests

PR #78, opened 2024-08-21
Add fuzzing + CI

### LINKED ISSUE #72: Fuzz compiler

uncomputable:
Fixes #72 

Implement fuzzing (`Arbitrary`) for the Simfony parse tree and write three fuzz targets:

1. `compile_text`: try to compile random ASCII strings; slow because of parser (#79)
2. `compile_parse_tree`: try to compile random parse trees; ~10x faster than `compile_text` and better coverage
3. `display_parse_tree`: check that `Display` → `ParseFromStr` is the identity function

Write a simple CI that uses nix flakes and dev shells. This CI is not more reproducible than regular GitHub CI, but it saves me time because I already have nix infrastructure in place. I will switch the CI to something reproducible in the near future.

Run the linter, unit tests (stable + MSRV) and the fuzzer in CI. For now, coverage is ignored in CI.


apoelstra:
Neat! I like this "budget" scheme for recursive types.

uncomputable:
First I had an increasing depth that is checked against a fixed maximum, but then I figured that a budget would be more flexible.

apoelstra (approved):
ACK fcc9c6f328752e2362dc9a80368fcd0933deee9e successfully ran local tests

PR #80, opened 2024-08-24
Update simplicity

uncomputable:
Update rust-simplicity to the latest version. Write documentation for the new jets. Fix documentation issues. Include codegen in CI.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -58,13 +58,25 @@ struct Scope {
     /// Later assignments occur higher in the tree than earlier assignments.
     /// ```
     variables: Vec<Vec<Pattern>>,
+    ctx: simplicity::types::Context,
 }
 
 impl Scope {
-    /// Create a new [`Scope`] for an `input` value that matches the pattern.
-    pub fn new(input: Pattern) -> Self {
+    /// Create the main scope.
+    ///
+    /// _This function should be called at the start of the compilation and then never again._
+    pub fn new() -> Self {
+        Self {
+            variables: vec![vec![Pattern::Ignore]],
+            ctx: simplicity::types::Context::new(),
+        }
+    }
+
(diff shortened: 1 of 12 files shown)


src/compile.rs
```diff
@@ -342,7 +352,11 @@ impl Call {
 /// The fold `(fold f)_n : E^(<2^n) × A → A`
 /// takes the list of type `E^(<2^n)` and an initial accumulator of type `A`,
 /// and it produces the final accumulator of type `A`.
-fn list_fold(bound: NonZeroPow2Usize, f: &ProgNode) -> Result<ProgNode, simplicity::types::Error> {
+fn list_fold(
+    ctx: &simplicity::types::Context,
+    bound: NonZeroPow2Usize,
+    f: &ProgNode,
+) -> Result<ProgNode, simplicity::types::Error> {
```
apoelstra, line 359: In 7368e4597e17d9528c5bff43746c02135ce38cc6:

Here and elsewhere you should be able to take the ctx out of `f` so you don't need an extra parameter and you won't risk calling the method incorrectly.

apoelstra:
3436b283fe5bc3a89b952ea618222113951e9aad looks good other than the ctx cleanups in the first commit.

src/compile.rs
uncomputable: Great catch. Thanks for pointing this out.

uncomputable:
Removed superfluous `ctx` function parameters.

apoelstra:
In 98d23965fa6d997dc53512ba11e0278429ee751b:

Need to also update the `codegen` crate which doesn't build on this commit.

uncomputable:
Merged the first two commits. Each commit should compile now.

apoelstra (approved):
ACK 0ccfa9b00395c1261f1a028d9c448b4c22c58c1d successfully ran local tests

PR #81, opened 2024-08-25
Fix asymptotics of BasePattern

### LINKED ISSUE #39: O(n^2) get method
[BasePattern::get](https://github.com/BlockstreamResearch/simfony/blob/0de1f8cf111369e86034d73e657922e951c510e4/src/scope.rs#L187-L207) has O(n^2) asymptotics although there exists an O(n) algorithm.

https://github.com/BlockstreamResearch/simfony/pull/38#discussion_r1616353544

> I would suggest using verbose_pre_order_iter in contains , and returning a vector of left/right decisions which you can pop every time you hit a BasePattern::Product in this loop. I'd also make contains non-public or rename it to reflect the richer return value.

### LINKED ISSUE #40: O(n^2) contains_all method
[BasePattern::contains_all](https://github.com/BlockstreamResearch/simfony/blob/0de1f8cf111369e86034d73e657922e951c510e4/src/scope.rs#L247-L252) has O(n^2) asymptotics although there exists a O(n) algorithm.

https://github.com/BlockstreamResearch/simfony/pull/38#discussion_r1616357299

> Instead you should construct a hashset of all the identifiers in self and then check that every item yielded from the iterator lands in the hashset.

uncomputable:
Depends on #80 _(because of the inference context parameters)_

Fixes #39 Fixes #40

### DIFF
--- a/src/pattern.rs
+++ b/src/pattern.rs
@@ -1,5 +1,5 @@
 use std::collections::hash_map::Entry;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::fmt;
 use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


apoelstra:
Can be rebased now. CI failure is "real" (cargo fmt). Nice!

uncomputable:
Rebased

src/pattern.rs
```diff
@@ -206,40 +206,31 @@ impl BasePattern {
         }
     }
 
-    /// Check if the `identifier` is contained inside the pattern.
-    pub fn contains(&self, identifier: &Identifier) -> bool {
```
apoelstra, line 210: In 25f810963b58cff14a7306edbd744b3c1d7595e8:

Might as well keep this method, though it can just be a oneliner `self.get(identifier).is_some()`.

apoelstra:
In 25f810963b58cff14a7306edbd744b3c1d7595e8:

The commit messages says the existing method was `O(2^n)` but you mean to say `O(n^2)` ... and really, `O(n * depth)`.

uncomputable:
Addressed the comments

apoelstra (approved):
ACK 8f3c94882d70c961ef38fdb82cbe0a9fefcafa94 successfully ran local tests

PR #82, opened 2024-08-31
Update simplicity

uncomputable:
Update rust-simplicity to the latest version. Type-check every Simfony value.


apoelstra (approved):
ACK c1068378ff4715b16f9c8099883bcc6a967abc74 successfully ran local tests; nice! Did not carefully check all the Value changes but very nice to see TypedValue go away

PR #83, opened 2024-09-02
Update examples

uncomputable:
Implement a harness for testing example programs. Add examples from Ivy and CashScript, and remove old examples.


apoelstra (approved):
ACK a9a6e26efbbe204932383465927c07bff41a4b43 successfully ran local tests; wow!

PR #84, opened 2024-09-02
Fix bitcoind tests

uncomputable:
Revive the bitcoind-tests sub crate for integration tests. Make it compile and check that it builds in CI to prevent future breaks. Integrate elementsd derivation in the nix flake.

This derivation takes a long time to build because it compiles elementsd from scratch. Once the derivation is built, it is cached, but I don't think we can use it for CI. Until we have some sort of caching, we have to run the integration tests locally. In any case, there is barely any integration test at the moment. I will add more tests once I can generate the witness data for them.


apoelstra (approved):
ACK 759a534a9df57ad5e6736556b0c02d22e2c7548c successfully ran local tests; just tested that the ordinary crate stuff works, did not test the flake

PR #85, opened 2024-09-09
QoL

uncomputable:
Some quality-of-life improvements. In particular, hex strings for arbitrary byte arrays, more constructors, and builtin aliases in witness json files.


apoelstra:
dc60ceeeb60d41edd91aa5f6ac20fd136e1f1c3e fails to build

```
error[E0599]: no variant or associated item named `ExpressionUnexpectedType` >
   --> src/value.rs:719:48
    |
719 |                         _ => return Err(Error::ExpressionUnexpectedType>
    |                                                ^^^^^^^^^^^^^^^^^^^^^^^^>
    |
   ::: src/error.rs:259:1
    |
259 | pub enum Error {
    | -------------- variant or associated item `ExpressionUnexpectedType` no>

error: aborting due to 1 previous error
```

but code-review-ack 8d93f95f23aa7e21ac2323a393957404e3b58054.

uncomputable:
I messed up during rebasing when I changed the order of commits. I wonder if we can include this check in CI...

apoelstra:
We could try this, but it's a bit of a PITA to get the list of commits in CI, and it'd take a long time to test every single commit (especially on 9-commit PRs like this).

apoelstra:
Alternately you could always run `git rebase -x 'cargo test' master` before pushing.

uncomputable:
@apoelstra Is the rest of the PR ok?

apoelstra:
@uncomputable yes

uncomputable:
I merged the PR without ACK :stuck_out_tongue:

PR #86, opened 2024-09-12
Reconstruct value

uncomputable:
Convert Simplicity values back to Simfony values via a Simfony type. We need this to display debug output from the Simplicity Bit Machine in terms of Simfony speak. Fix bugs that came up during fuzzing.

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -24,6 +24,10 @@ impl<'a, A: Clone> BTreeSlice<'a, A> {
     /// Fold the tree in post order, using the binary function `f`.
     ///
     /// Returns `None` if the tree is empty.
+    ///
+    /// ## See
+    ///
+    /// Opposite of [`Unfolder::unfold`].
     pub fn fold<F>(self, f: F) -> Option<A>
     where
         F: Fn(A, A) -> A,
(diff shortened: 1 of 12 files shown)


src/value.rs
```diff
@@ -848,7 +848,7 @@ impl ValueConstructible for StructuralValue {
         bound: NonZeroPow2Usize,
     ) -> Self {
         let elements: Vec<Self> = elements.into_iter().collect();
-        assert!(bound.get() <= elements.len(), "Too many elements");
+        assert!(elements.len() < bound.get(), "Too many elements");
```
apoelstra, line 899: In 3d8a00854e31f88573df7687e6187d99c231dacc:

I struggled to understand this commit. If you mean to assert that there are strictly fewer elements than the bound (which does sound right) then you need to change the error message in the `assert`.

The commit message also says that this code is called "after another assertion that checks the correct invariant", which I *think* refers to line 746 which `debug_assert`s equality between `elements.len` and some other quantity. But I don't see what that assertion has to do with this one.

Maybe you are referring to the assertion in `Partition::from_slice`? That assertion appears to be equivalent to this new one, which means that the previous code should never have been possible to call! What's going on here?

uncomputable: `<StructuralValue as ValueConstructible>::list` is called by `<StructuralValue as From<&'a Value>::from`. The latter has the assertion `debug_assert!(elements.len() < bound.get())`. I will update the assert message. "Too many elements" is meant to be displayed for the error case, but I can see how a statement about the should-be state is clearer.

apoelstra: I find with assert messages it's clearest to use "should" language. "There should be strictly fewer elements than the bound". Also in this case we should put the exact numbers into the error messages (since the `assert!` macro can't provide them itself as it's just asserting a boolean).

uncomputable:
Clarified the assertion about list bounds.

uncomputable:
Included the bound value in the assert message, for easier debugging.

src/value.rs
apoelstra: Cool, I'm happy with this. Though still a little mystified that the original code ever worked. I think it must have never been called.

src/array.rs
```diff
@@ -72,18 +72,19 @@ impl<'a, A: Clone> TreeLike for BTreeSlice<'a, A> {
 
 /// Partition of a slice into blocks of (lengths of) powers of two.
 ///
-/// The blocks start at (length) `N` and decrease to one in order.
-/// Depending on the (length of the) slice, some blocks might be empty.
+/// ## Partition of less than 2^(n + 1) elements
```
apoelstra, line 129: In 7f178c43b0595e09eb099228224d41556342ea0f:

Should not have the `##` and the first line should end with a `.`.

apoelstra: Oh never mind, I am misreading the diff.

apoelstra:
Lots of cool ideas in here! In particular generating arbitrary values by first choosing an arbitrary type and then choosing an arbitrary value of that type. I am really tempted to try generating these using [Boltzmann generators](https://algo.inria.fr/flajolet/Publications/DuFlLoSc04.pdf) by which we would choose a target size (in bits) and then could uniformly generate an object of that size. (Much of the complexity of this approach is in obtaining uniformity, but as a side-effect we don't need to "know the structure up front" beyond its size. In practice this is probably what we want! The exact structure of `Arbitrary`-generated objects we don't really care about, and we struggle to bound their size!)

Another relevant concept is that of an [anamorphism](https://blog.sumtypeofway.com/posts/recursion-schemes-part-2.html) which talks about recursively generating objects as an inverse-fold operation. My CS fundamentals are not strong enough to verify that your `Unfolder` aligns with the concepts in that blog post, but I guess one of us ought to study this harder.

Upstream in rust-simplicity we want to eventually replace the recursive type with a flat one, which should greatly speed up the fuzztest (and other parts of this code).

Anyway code review ack. Let me run my local CI and I will give you a real ack.

uncomputable:
It took me some time to bring the PR into shape. Fuzzing helped with finding bugs. We might want to test parts of the code with property tests instead of fuzzing. Restricting the maximum depth of a generated value already leans towards this direction, I feel.

apoelstra:
In 6ffe45cb:

There is a clippy lint on line 771:

```
error: redundant closure
   --> src/value.rs:771:59
    |
771 |                 1 => Self::arbitrary_of_type(u, ty_r).map(|val| Self::some(val)),
    |                                                           ^^^^^^^^^^^^^^^^^^^^^ help: replace the closure with the function itself: `Self::some`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure
    = note: `-D clippy::redundant-closure` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::redundant_closure)]`
```

uncomputable:
Ah, my CI doesn't yet check all feature combinations. I pushed a fix and I will update my local scripts.

apoelstra (approved):
ACK 88d15d3af847741aa08b107ebcfdc263d5c2ad18 successfully ran local tests

PR #87, opened 2024-09-16
Add debug symbols

uncomputable:
Track fallible Simfony expressions via debug symbols that are included in the Simplicity target code (more details are in the commit message).

Add public methods to convert debug symbols into helpful error messages by providing the values that the Bit Machine sees at runtime. Debug symbols cover all failure cases, so the Bit Machine should always be able to display a helpful Simfony error.

Add `dbg!` expressions that can be used print their input value at runtime.


apoelstra:
code review ack 88d15d3af847741aa08b107ebcfdc263d5c2ad18

CI failure looks real (I'll wait til it's green to run my local CI).

Super cool concept!

PR #91, opened 2024-09-30
Witness update

uncomputable:
Add some methods to make the map of witness values more useful. Test witness serialization in a fuzz test. Fix some docs.


src/witness.rs
```diff
@@ -218,6 +218,25 @@ impl<'de> Deserialize<'de> for WitnessName {
...
+
+        let mut map = serializer.serialize_map(Some(self.len()))?;
+        for (name, value) in self.iter() {
+            let value_map = BTreeMap::from([
+                ("value", value.to_string()),
+                ("type", value.ty().to_string()),
+            ]);
+            map.serialize_entry(name.as_inner(), &value_map)?;
+        }
+        map.end()
```
apoelstra, line 237: In 5d7a91a4dd9b21a98e3b9f034c18ff9eb5a723d2:

Heh, there is definitely a more efficient way to do this than to actually construct a `BTreeMap` and serialize it. But I wouldn't worry about it for now. It's an opportunity for future optimization.

apoelstra (approved):
ACK 70535b6cbc7456a988e80f32bc478ecf82bc8844 successfully ran local tests

PR #93, opened 2024-10-08
Add witness module

uncomputable:
Make witness JSON parsing via serde optional. Introduce witness modules as an alternative way to provide witness data. Witness modules are written directly inside the Simfony source code.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,7 +70,8 @@ jobs:
           - compile_text
           - display_parse_tree
           - parse_value_rtt
-          - parse_witness_rtt
+          - parse_witness_json_rtt
+          - parse_witness_module_rtt
           - reconstruct_value
     steps:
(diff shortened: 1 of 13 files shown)


apoelstra:
Neat! Concept ACK. Though I imagine in the future this will become a weird corner of the language that is only really used by the web IDE.

uncomputable:
I moved most changes to #94. The fuzzer bug has to do with the witness module, which kind of lives inside the parse tree and kind of doesn't.

uncomputable:
I fixed the fuzzer.

Writing witness data directly into the program is a huge feature for the IDE, but I agree that it feels kind of weird. Keeping the program parser separate from the witness parser will allow us to remove the feature if we don't need it.

apoelstra (approved):
ACK e3ed9039589af23691e6f41591cedb3284c85b48 successfully ran local tests

PR #94, opened 2024-10-08
Optional serde

uncomputable:
Make witness JSON parsing via serde optional. Add a fuzz target for parsing const expressions as values. Update CI to run all fuzz targets in parallel.


apoelstra (approved):
ACK 6f1389882e958bf8f087617c8c3c4c75b5c8e5ca successfully ran local tests

PR #95, opened 2024-10-10
Document jet cost

uncomputable:
Simfony-as-Rust is the best resource on jets at the moment and it might be useful to know how expensive a jet will be on the blockchain.


apoelstra (approved):
ACK ad3aae8c702a7e99353785cf14b1250e8d16b4d2 successfully ran local tests

PR #96, opened 2024-10-11
Update simplicity

uncomputable:
Update rust-simplicity to the version that runs on Liquid testnet.


apoelstra (approved):
ACK 44d23e09dc1074309be5d343f06c29fe08cd6b40 successfully ran local tests

PR #97, opened 2024-10-12
Fix integration tests

uncomputable:
Bring the integration tests up to speed with Liquid testnet. Fix a bug that the fuzzer found.

### DIFF
--- a/bitcoind-tests/elementsd-simplicity.nix
+++ b/bitcoind-tests/elementsd-simplicity.nix
@@ -1,12 +1,12 @@
 { pkgs
 }:
 pkgs.elementsd.overrideAttrs (_: {
-  version = "unstable-2024-09-02";
+  version = "liquid-testnet-2024-10-08";
   src = pkgs.fetchFromGitHub {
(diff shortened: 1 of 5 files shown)


apoelstra:
Nice! Will test and ACK.

Unfortunately I can't test the integration tests locally unless the integration test crate were part of the workspace and if the lockfiles were polluted with all the integration test deps (which is part of why the integration tests are *not* part of the workspace..).

It's easy for me to patch Cargo.toml before testing to add stuff to the workspace, but not so easy to patch the lockfiles :(. This is something I might be able to address when I migrate to flakes, since I need to do some significant rearchitecting of my configuration when I do that anyway.

uncomputable:
The integration tests are pretty expensive to run. In particular, you have to build elementsd from source :/ When I flesh out the integration tests more, I will think of ways to integrate them into CI. This should also help you with your local CI.

apoelstra (approved):
ACK 5a7c2b6fb24c740b674caffd2f72ffa1df58a8c6 successfully ran local tests

PR #99, opened 2024-10-20
Fix WASM

uncomputable:
This crate was broken on WASM; make it compile again. Check WASM compilation in CI. Fix broken debug symbols on WASM.


uncomputable:
Fixed how the value of `dbg!(x)` is printed when using debug symbols.

uncomputable:
See https://github.com/BlockstreamResearch/rust-secp256k1-zkp/pull/87

uncomputable:
Updated this PR to use the latest version of rust-secp256k1-sys. Remove the Cargo.toml `[patch]`.

apoelstra (approved):
ACK 7c91cab9958653596e60b363d6bf32448a33222a; successfully ran local tests

PR #100, opened 2024-11-05
Parameterize programs

uncomputable:
Refactor the code that handles witness types / values / modules. Add parameterized values to Simfony which are instantiated at commitment time.

Simfony programs with parameters are templates for a family of Simfony programs. A template is instantiated by providing arguments to its parameters. Arguments can be provided via JSON or via code `mod param { ... }`.

The existing API needed to be updated to handle parameters. This turned out to be more work than expected :) While most changes are mechanical, there were some interesting engineering problems, such as handling parameters inside fuzz tests.

### DIFF
--- a/bitcoind-tests/tests/common/daemon.rs
+++ b/bitcoind-tests/tests/common/daemon.rs
@@ -1,11 +1,11 @@
 use std::str::FromStr;
 
-use simfony::elements;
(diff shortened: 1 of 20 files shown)


apoelstra (approved):
ACK c5ea02a9480682bb221c542b6290b93b6a9dea9f; successfully ran local tests; yeah, this seems like a good direction

PR #101, opened 2024-11-26
Fix display

uncomputable:
Fixes how modules and values are displayed.


apoelstra (approved):
ACK 66d44621622f3510e5a28e1d9e93be8b0089dd9b; successfully ran local tests

PR #103, opened 2024-12-09
Simfony book

uncomputable:
Write the first draft of the Simfony book. The book will be updated in following PRs. Add a GitHub action to deploy the book in GitHub pages.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -1,0 +1,31 @@
+name: Deploy Book on GitHub Pages
+on:
+  push:
+    branches:
+      - master
+permissions:
+  contents: write
+jobs:
+  deploy:
+    name: Deploy Book on GitHub Pages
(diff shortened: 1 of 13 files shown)


uncomputable:
I clarified the above mentioned section.

`Some(42)` can only be pattern-matched one level deep.

```rust
let unwrap_or_default: u32 = match Some(42) {
    None => 0,
    Some(x) => x,
};
```

However, in Rust, `Some((4, 2))` can be matched two levels deep. This is not possible in Simfony.

```rust
let unwrap_or_default: u32 = match Some((4u16, 2u16)) {
    None => 0,
    Some((y, z)) => ((y as u32) << 16) + z as u32,
};
```

apoelstra (approved):
ACK a0379a859580fe4fe9a43df1a5faef79f10f384f; successfully ran local tests; much better!

PR #109, opened 2024-12-23
Add Non interactive fee bumping example

sanket1729:
Had to refactor some of the dummy_env stuff.

If I try to include change output in the default env, everything breaks and I did not want to spend time trying to fix all of tests. Instead, we create a new default with fee just for this one test case.

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -20,7 +20,7 @@ impl<'a, A> BTreeSlice<'a, A> {
     }
 }
 
-impl<'a, A: Clone> BTreeSlice<'a, A> {
+impl<A: Clone> BTreeSlice<'_, A> {
     /// Fold the tree in post order, using the binary function `f`.
     ///
     /// Returns `None` if the tree is empty.
@@ -57,7 +57,7 @@ impl<'a, A: Clone> BTreeSlice<'a, A> {
     }
 }
 
-impl<'a, A: Clone> TreeLike for BTreeSlice<'a, A> {
+impl<A: Clone> TreeLike for BTreeSlice<'_, A> {
     fn as_node(&self) -> Tree<Self> {
         match self.0.len() {
             0 | 1 => Tree::Nullary,
@@ -169,7 +169,7 @@ impl<'a, A> Partition<'a, A> {
     }
 }
 
-impl<'a, A: Clone> Partition<'a, A> {
+impl<A: Clone> Partition<'_, A> {
     /// Check if the partition is complete.
     ///
     /// A complete partition contains no empty blocks.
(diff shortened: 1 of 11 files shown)


apoelstra:
In 8231b9547fb1c064321df154fb9ed6b44a13468e: trailing whitespace on several lines.

apoelstra:
In 8231b9547fb1c064321df154fb9ed6b44a13468e:

Looks like you're using `left_pad` in a couple places to widen integer types. You can just use `<u64>::from` or `<u32>::into`.

Your changes to the test environment seem fine to me.

apoelstra:
Done reviewing 16ebe7411e78dd5e5d02b67d6ebc083976cd7dfd

apoelstra:
cc @uncomputable can you provide some guidance here? It doesn't seem right that the user has to call a jet explicitly for integer promotion.

sanket1729:
```
let v2 : u64 = <u1>::into(v1);
   |                    ^^^^^^^^^^^^^^ Cannot cast values of type `u1` as values of type `u64`
```

src/array.rs
```diff
@@ -20,7 +20,7 @@ impl<'a, A> BTreeSlice<'a, A> {
     }
 }
 
-impl<'a, A: Clone> BTreeSlice<'a, A> {
+impl<A: Clone> BTreeSlice<'_, A> {
```
uncomputable, line 23: ab42d67c78505463cf66b7f54a4137ea017e51d4: Are you using rust analyzer? `clippy` didn't show me these lints.

uncomputable:
Casting is possible when two Simfony types are based on the same Simplicity type. Casting is a NOP in terms of Simplicity. Integers with different bit width are different types in Simplicity, so there is no cast.

To convert `u1` to `u64`, [`left_pad_low_1_64` can be called](https://docs.rs/simfony-as-rust/latest/simfony_as_rust/jet/multi_bit_logic/fn.left_pad_low_1_64.html). That seems to be the cheapest and most concise way. Alternatively, a nested tuple of integers and Booleans can be cast into `u64`. The cast itself is free, but constructing the tuple is not.

```
let b: bool = true;
let x: u64 = <(u32, (u16, (u8, (u4, (u2, (u1, bool))))))>::into((0, (0, (0, (0, (0, (0, b)))))));
assert!(jet::eq_64(x, 1));
```

uncomputable (approved):
ACK ab42d67c78505463cf66b7f54a4137ea017e51d4 Feel free to optimize the `check_neg`  function.

uncomputable:
I have code to dynamically generate witness data for most of the Simfony examples, so we can change the Elements environment without every test failing. I will make a PR when the API is ready and integrate the hotfix from this PR.

apoelstra (approved):
ACK ab42d67c78505463cf66b7f54a4137ea017e51d4; successfully ran local tests

sanket1729:
I am planning to post this example on delving bitcoin to get some more attention to simphony and some cool applications. Let me know if you have any objections

uncomputable:
That's a great idea. Make sure you hit the right spelling :)

PR #111, opened 2025-01-30
Update simplicity

uncomputable:
Update rust-simplicity for pruning support. Update the nix flake.


uncomputable:
Updated to NixOS 24.11

PR #112, opened 2025-01-30
More sighashes

uncomputable:
Add examples for `SIGHASH_SINGLE`, `SIGHASH_ALL | SIGHASH_ANYPREVOUT` and `SIGHASH_ALL | SIGHASH_ANYONECANPAY`.


apoelstra:
These examples look great!

Do these match the actual sighash modes used in non-Simplicity Taproot?

I wonder if it'd be worthwhile to add a "meta-example" which takes a 65-byte signature, looks at the sighash flag at the end, and then chooses the right verification function based on that.

apoelstra:
Yes, I understand that. But here you have written large blocks of Simfony code which manually hashes parts of the transaction. My question was

> Do these match the actual sighash modes used in non-Simplicity Taproot?

and I think we should add the answer to this question to the docs somewhere.

apoelstra (approved):
ACK 7602e8ad7cc177398f539010bacbd089021b5902; successfully ran local tests

PR #113, opened 2025-02-25
Mark const methods

uncomputable:
Mark some key methods as const.


apoelstra (approved):
ACK c0c62db3749c206a37af5213f16896f948c59acc; successfully ran local tests

PR #125, opened 2025-04-29
Satisfy program with optional pruning

m-kus:
This PR exposes an API for satisfying a program with pruning given an optional environment.  
The naming might be a bit misleading (maybe `satisfy_pruned` would be better?), pls let me know if you want to change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,6 +22,7 @@ mod witness;
 
 use std::sync::Arc;
 
+use simplicity::jet::elements::ElementsEnv;
 use simplicity::{jet::Elements, CommitNode, RedeemNode};
 
 pub extern crate either;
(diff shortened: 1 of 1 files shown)


apoelstra:
Would be good to move away from the stringly-typed errors, but I guess that's independent of this PR.

I'm gonna go ahead and ACK this -- I agree the name is a bit weird but the whole API right now is a bit undiscoverable and I think we want to do a pass through it and see if we can't reduce the number of exposed types and clarify their relationship.

src/lib.rs
```diff
@@ -133,15 +134,31 @@ impl CompiledProgram {
...
+        self.satisfy_with_env(witness_values, None)
+    }
+
+    /// Satisfy the Simfony program with the given `witness_values`.
+    /// If `env` is `None`, the program is not pruned, otherwise it is pruned with the given environment.
+    ///
+    /// ## Errors
+    ///
+    /// - Witness values have a different type than declared in the Simfony program.
+    /// - There are missing witness values.
```
uncomputable, line 146: 30f125f18942a096edef794df7c0c06ea42345bd: nit: pruning fails if the program doesn't run successfully for the given environment and witness data.

uncomputable (approved):
ACK 30f125f18942a096edef794df7c0c06ea42345bd

apoelstra (approved):
ACK 30f125f18942a096edef794df7c0c06ea42345bd; successfully ran local tests

PR #127, opened 2025-04-29
Turn on a bunch of cippy ints

apoelstra:
This is the first part of a project to turn on a ton of pedantic Clippy lints. We can choose not to do some if they're too annoying. There are about 20 left after this.


apoelstra (commented):
On 8e15fd3e032978584b5839100a29cba568eece73 successfully ran local tests

apoelstra:
cc @canndrew can you take a look at this one?

apoelstra:
And @uncomputable since this is kinda your crate and I'm making style changes.

uncomputable (approved):
ACK 8e15fd3e032978584b5839100a29cba568eece73 Looks good to me

clippy.toml
```diff
@@ -1 +1,3 @@
 msrv = "1.78.0"
+# We have an error type, `RichError`, of size 144. This is pushing it but probably fine.
+large-error-threshold = 145
```
canndrew, line 3: Just boxing the `Error` in `RichError` would cut down the size a fair bit. I was gonna do a PR to replace the `String` errors with `anyhow::Error` in a few places, so I'll include it in that.

apoelstra: That sounds good. If you can remove this clippy.toml line I'd appreciate it!

Also, because this is both a library and a CLI tool, it may be a judgement call where to use `anyhow` ... I'd prefer we not use it on any "library errors" and only on "application errors".

Historically we've manually done all our errors, which is a bit tedious. I wouldn't mind if you wanted to introduce `thiserror` here. In rust-simplicity probably not, because we're much more agressive about pruning dependencies there.

canndrew: Yeah anyhow is kind of a lazy hack. Just less so than using `String`. I agree that libraries should use proper errors that you can actually match on. I'll fix up the PR I just opened.

PR #129, opened 2025-05-05
Use clap for parsing CLI args

canndrew:
Add clap as a dependency and use it to parse the command line arguments. Also add a --debug cli option to include debug symbols.

PR #130, opened 2025-05-05
Use anyhow::Error in place of String

canndrew:
Use the anyhow crate for errors where we're currently just using String. This makes it easier to add context to errors and avoids the need to convert to a string when bubbling up errors.

Also box the `Error` field in `RichError` to reduce its size and keep clippy happy.

This PR builds off of #129.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,6 +62,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


canndrew:
I'll fix the changes on the library side to use a new error type rather than anyhow then re-open this.

PR #131, opened 2025-07-14
Rename from Simfony to SimplicityHL

kiarafbickers:
Changed all mentions of Simfony to SimplicityHL in docs, comments, urls, and text. No code, filenames, or logic was changed, just text!


uncomputable:
The crate needs to be renamed on crates.io: https://crates.io/crates/simfony

PR #133, opened 2025-07-28
Remove references to Simfony

imaginator:
AFAIK the only remaining reference is to  https://github.com/uncomputable/simfony-as-rust (generates: https://docs.rs/simfony-as-rust/) in the file `SimplicityHL/book/src/function.md`. I would presume that this should be updated too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -219,7 +219,7 @@ checksum = "f46ad14479a25103f283c0f10005961cf086d8dc42205bb44c46ac563475dca6"
 name = "codegen"
 version = "0.1.0"
 dependencies = [
- "simfony",
+ "simplicityhl",
 ]
 
 [[package]]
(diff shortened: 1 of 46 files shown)


apoelstra:
Welcome @imaginator!

In your second commit it looks like you autoformatted everything with the stable formatter. Unfortunately this is missing a bunch of options that we use. Can you run `cargo +nightly fmt` to reformat it with the nightly rules?

imaginator:
good to know about `cargo +nightly fmt`. I spent ages trying to recreate the formatting with a `rustfmt.toml` file until I gave up and redid everything manually - but three files slipped though.

apoelstra (approved):
ACK 4493b607788a85ab4042d1c603299a93a2cca3ed; successfully ran local tests; thanks!

PR #135, opened 2025-08-06
remove orphan heading "Locking and unlocking UTXOs"

imaginator:
While I'd love to know what was going to be added here, it's better that we remove the headline until we have content for it.


uncomputable (approved):
ACK f693c16e9dc26d47543723d31a7650545e4162b2

PR #143, opened 2025-09-12
refactors to prepare for the new rust-simplicity version

apoelstra:
Since https://github.com/BlockstreamResearch/rust-simplicity/pull/305 we are not allowed to have type inference contexts (and therefore `ConstructNode`s) that escape the lexical scope that defines them.

We currently use `ConstructNode` inside `ProgramNode`. This PR changes this to `CommitNode`, which is more correct, as well as being necessary to avoid exposing bad `GhostCell` ergonomics to users of this library.

### DIFF
--- a/src/compile/builtins.rs
+++ b/src/compile/builtins.rs
@@ -2,7 +2,8 @@ use std::num::NonZeroUsize;
 
 use simplicity::node::CoreConstructible;
 
-use crate::{named::CoreExt, ProgNode};
+use super::ProgNode;
+use crate::named::CoreExt;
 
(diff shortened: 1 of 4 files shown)


apoelstra:
cc @canndrew can you take a look at this one?

delta1 (approved):
ACK 96e7df240a5ed0fa60b3dcab5e2d9b16dcd9fe1c

Ran cargo test locally on rust stable 1.90

delta1 (approved):
ACK 760c51cf2b7db48824fc59c6774d2c8df4a60a7e 

Ran cargo test successfully

Issue #147, opened 2025-09-26
VSCode extention with LSP support


gerau:
We are currently working on the [SimplicityHL Language Server](https://github.com/distributed-lab/simplicityhl-lsp)  (which mentioned in [this pull request](https://github.com/BlockstreamResearch/SimplicityHL/pull/146#issuecomment-3334668606)).

At the moment, language server provides two main features:

- Error diagnostics
![diagnostics](https://github.com/user-attachments/assets/54315645-464b-40c3-bb72-c6e8c4bc0ad5)
- Completion for functions and jets
![completion](https://github.com/user-attachments/assets/bbc2b9de-c286-4d31-b47e-ac95885f8916)

We plan to expand the functionality for working with SimplicityHL in future updates, and the language server will be the main development tool we will be working on.

Can we merge our VSCode extension functionality for the LSP into your official extension?

apoelstra:
>Can we merge our VSCode extension functionality for the LSP into your official extension?

Yes, please!

PR #148, opened 2025-09-29
VSCode extension with LSP support

gerau:
This pull request introduces optional support for the SimplicityHL language server, which was discussed in #147. 
Essentially, it works like this:

- When opening a `.simf` or `.wit` file, the language server starts automatically.
- If the language server is not found in `PATH` or other common locations, the extension offers to install it.
- Users may dismiss this notification by selecting **“Don't show again”**, after which it will no longer appear.

Also, some additional changes have been made:
- Updated the **README** to describe and showcase the capabilities of the language server.
- Changed the language identifier in `package.json` from `simfony` to `simplicityhl` for both `.simf` and `.wit` files, and added the required dependencies.
- Added `esbuild.js` to bundle Typescript files.

### DIFF
--- a/vscode/.vscodeignore
+++ b/vscode/.vscodeignore
@@ -3,7 +3,7 @@ node_modules/
 .git/
 .gitignore
 npm-debug.log
-*.ts
+**/*.ts
 *.map
 tsconfig.json
 .github
--- a/vscode/README.md
+++ b/vscode/README.md
@@ -10,8 +10,16 @@ SimplicityHL is a high-level language for writing Simplicity smart contracts. Si
 
 - Syntax highlighting for .simf and .wit files
 - Basic language configuration (brackets, comments)
+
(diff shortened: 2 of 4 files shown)


apoelstra:
Can you please restructure this PR so that each commit is independent and self-contained, and there are no merge commits?

apoelstra:
Thanks! Though I wonder why b59ed7e9735923c2eab1691782f941e354b0e5af makes a ton of indentation changes to package-lock.json that 1c11594c1eea3602cf679d083a20a9a2cc5a65d3 largely reverts.

Anyway, I will go ahead and ACK and merged this. Confirmed that it doesn't touch any real code and that it looks mostly harmless. We will reconsider moving it to its own repo if there is a lot of activity of this form.

apoelstra (approved):
ACK 1c11594c1eea3602cf679d083a20a9a2cc5a65d3; successfully ran local tests

PR #156, opened 2025-10-24
Switch to BIP-0341's NUMS key

roconnor-blockstream:


apoelstra (approved):
ACK 2b872a7ce56b356cccbb935c91c1e4d7312695a0; successfully ran local tests

PR #168, opened 2025-11-17
Add "Restart Server" command to VSCode extension

gerau:
This is a simple quality-of-life feature for the language server. Currently, the only way to restart the server is to reload the entire VSCode window, which is not convenient, especially for testing.


imaginator (approved):
This looks great to me. Thanks for taking care of this.

apoelstra:
https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/main/github-merge.py

PR #172, opened 2025-11-19
LSP: refactor completions, add specific type cast completions

gerau:


lsp/src/completion/tokens.rs
```diff
@@ -0,0 +1,65 @@
...
+    DoubleColon,
+    OpenAngle,
+    CloseAngle,
+    EqualSign,
+    OpenBracket,
+    ClosedBracket,
+    Identifier(String),
+    Jet,
+}
+
```
KyrylR: For every parsing function please include string that is going to be accepted by the function

lsp/src/completion/mod.rs
```diff
@@ -73,6 +81,68 @@ impl CompletionProvider {
...
+            Err(_) => return None,
+        };
+
+        match tokens.as_slice() {
+            [Token::Jet, ..] => Some(self.jets.clone()),
+
+            // Case for ": type = <", so we can return completion for specific type, or generic one
+            // if it is not on default type casts.
+            [Token::OpenAngle, Token::EqualSign, Token::Identifier(type_name), Token::Colon, ..]
+            | [Token::Identifier(_) | Token::OpenBracket, Token::OpenAngle, Token::EqualSign, Token::Identifier(type_name), Token::Colon, ..] =>
```
KyrylR: This should be handled by the nom

KyrylR (changes requested):

KyrylR:
ACK 7d4e59d1ba06bd6545b00db977701caf0ad15acc

Ran tests locally and tested new features in VSCode

Issue #179, opened 2025-12-09
Make simc output more machine-readable


schoen:
### Project

compiler

### Describe the feature

Right now, when you compile a SimplicityHL file, you get textual output with either 2 or 4 lines, where the base64-encoded compiled program appears on line 2 and the serialized witness, if present, appears on line 4.

This isn't exactly *hard* to parse, but it feels kind of awkward to be specifying line numbers of output in order to retrieve these objects. Could we have an option to save these into files? Or an option for JSON output that we could then parse with a JSON parser, like with much of the `hal-simplicity` output?

PR #180, opened 2025-12-09
feature(simc): flag for json output

delta1:
re: #179

Adds a `--json` flag to simc to output the resulting program/witness as JSON


apoelstra (approved):
ACK 55e7fd86105b1ae108d482d5323c6c5078c804cc; successfully ran local tests; though we should really rewrite and refactor this main function sometime..

Issue #79, opened 2024-08-21
Parsing is slow


uncomputable:
The PEST parser struggles with series of open brackets that are not closed. This means that the fuzzer runs slow units that take up to hundreds of seconds to parse.

```rust
fn n(){ { (s,(( (Ns,(s,(x,(((s,((s,(s,(s,(x,(( {5
```

```rust
fn fnnfn(MMet:(((sssss,((((((sssss,ssssss,ss,((((((sssss,ss,((((((sssss,ssssss,ss,((((((sssss,ssssss,((((((sssss,sssssssss,(((((((sssss,sssssssss,(((((ssss,((((((sssss,sssssssss,(((((((sssss,ssss,((((((sssss,ss,((((((sssss,ssssss,ss,((((((sssss,ssssss,((((((sssss,sssssssss,(((((((sssss,sssssssss,(((((ssss,((((((sssss,sssssssss,(((((((sssss,sssssssssssss,(((((((((((u|(
```

```rust
fn nf224(jet:J ){{{{{{{{{{{{ {     {{ {  { {   {{ 
```

Maybe the grammar file that I wrote contains anti patterns that needlessly slow down the parser. So far I haven't found a resource on these, but I will keep looking and I will try to rewrite the file.

It could also be that the PEST parser is inherently slow. In this case, we will have to move to a different parser. However, I will hold off on such a massive change until more important work has been done.

PR #185, opened 2025-12-18
pest to chumsky migration

gerau:


gerau:
Right now there is a working parser using the `chumsky` crate which replicates the behavior of the `pest` parser in terms of building a correct parse tree -- it should produce the same Simplicity program. This implementation also fixes #79. 

Error reporting is currently broken because we need to replace the logic of `parse::ParseFromStr` to return multiple errors or handle recoverable errors differently, and error recovery is proving to be more overwhelming than I estimated it would be.

The code will be refactored because some parts are only half-finished (such as adding `Spanned` for certain names) and there are better ways to use parser combinators. However, I want to show this progress before implementing error recovery.

src/lib.rs
```diff
@@ -632,7 +632,6 @@ fn main() {
     }
 
     #[test]
-    #[ignore]
```
uncomputable: 1b1e7513a2d08f808113b0af0c9067c24a10c32e It's nice to see that chumsky seems to be faster than pest here.

src/error.rs
```diff
@@ -1,155 +1,112 @@
...
     }
 
     /// Return a slice from the given `file` that corresponds to the span.
-    ///
-    /// Return `None` if the span runs out of bounds.
     pub fn to_slice<'a>(&self, file: &'a str) -> Option<&'a str> {
-        let mut current_line = 1;
-        let mut current_col = 1;
-        let mut start_index = None;
+        if file.is_empty() && self.start == 0 && self.end == 0 {
```
canndrew: You can just do `file.get(self.start..self.end)` here. That'll also handle indexes into the middle of multi-byte codepoints without panicking.

src/error.rs
```diff
@@ -248,15 +213,25 @@ impl fmt::Display for RichError {
...
 
-                let (underline_start, underline_length) = match self.span.is_multiline() {
+                let line_start_byte = index
+                    .offset(LineCol {
+                        line: start_pos.line,
+                        col: 0,
+                    })
+                    .map_or(0, |ts| u32::from(ts) as usize);
+
+                let start_col = file[line_start_byte..self.span.start].chars().count();
```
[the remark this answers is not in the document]
canndrew: Actually I just checked and LSP now allows you to choose between utf{8,16,32} at your leisure. But it's moot anyway since this is just deciding how long an underline to print and that's going to depend on the terminal.

gerau:
Also a note about performance: `chumsky` seems faster in general than `pest` parser. For example, on my machine for a large file `.simf` file, which was generated by  [simplicity-bn254](https://github.com/distributed-lab/simplicity-bn254), `chumsky` is 10 times faster than `pest` for parsing. But trade-off for this is slower compilation times and lag with `rust-analyzer`, because `chumsky` is type-driven.
 
It would be nice if we could move the parser to a different crate, so it would not affect compile time too much, and the SimplicityHL parser could be used separately from the compiler.

src/error.rs
```diff
@@ -1,155 +1,112 @@
...
     pub fn cmr(&self) -> Cmr {
         let mut hasher = sha256::HashEngine::default();
-        hasher.input(&self.start.line.get().to_be_bytes());
-        hasher.input(&self.start.col.get().to_be_bytes());
-        hasher.input(&self.end.line.get().to_be_bytes());
-        hasher.input(&self.end.col.get().to_be_bytes());
+        hasher.input(&self.start.to_be_bytes());
+        hasher.input(&self.end.to_be_bytes());
         let hash = sha256::Hash::from_engine(hasher);
         Cmr::from_byte_array(hash.to_byte_array())
```
KyrylR: If this function is not used at all, maybe let's just remove it?

src/lib.rs
```diff
@@ -52,12 +53,17 @@ impl TemplateProgram {
...
+        let parse_program = parse::Program::parse_from_str_with_errors(&file, &mut error_handler);
+        if let Some(program) = parse_program {
+            let ast_program = ast::Program::analyze(&program).with_file(Arc::clone(&file))?;
+            Ok(Self {
+                simfony: ast_program,
+                file,
+            })
+        } else {
+            Err(error_handler)?
+        }
```
KyrylR: Is it possible to have same error type for both?

src/parse.rs
```diff
@@ -862,762 +860,1000 @@ impl_parse_wrapped_string!(WitnessName, witness_name);
...
+                    "u1" => AliasedType::u1(),
+                    "u2" =>  AliasedType::u2(),
+                    "u4" =>  AliasedType::u4(),
+                    "u8" => AliasedType::u8(),
+                    "u16" => AliasedType::u16(),
+                    "u32" => AliasedType::u32(),
+                    "u64" => AliasedType::u64(),
+                    "u128" => AliasedType::u128(),
+                    "u256" => AliasedType::u256(),
+                    "Ctx8" | "Pubkey" | "Message64" | "Message" | "Signature" | "Scalar" | "Fe" | "Gej"
```
KyrylR: We could mention those in an issue about hardcoding as well

src/parse.rs
KyrylR: No unit tests is quite troublesome, though this file is already huge

src/error.rs
```diff
@@ -235,9 +229,17 @@ impl fmt::Display for RichError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self.file {
             Some(ref file) if !file.is_empty() => {
-                let start_line_index = self.span.start.line.get() - 1;
-                let n_spanned_lines = self.span.end.line.get() - start_line_index;
-                let line_num_width = self.span.end.line.get().to_string().len();
+                let index = LineIndex::new(file);
+
+                let start_pos = index.line_col(TextSize::from(self.span.start as u32));
+                let end_pos = index.line_col(TextSize::from(self.span.end as u32));
```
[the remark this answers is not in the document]
gerau: `Span` has a start and an end, so the only way to implement the `From` trait would be for a tuple of `TextSize`s. That isn't a very pretty solution either, so I'd prefer to stick with the current variant.

src/lexer.rs
```diff
@@ -0,0 +1,206 @@
...
+    // Boolean type
+    BooleanType,
+
+    // Built-in functions
+    Macro(&'src str),
+
+    // Comments and block comments
+    //
+    // We would discard them for the compiler, but they are needed, for example, for the formatter.
+    Comment,
```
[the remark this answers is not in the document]
KyrylR: Did you create an issue?

src/parse.rs
KyrylR: Did you create an issue?

src/lib.rs
```diff
@@ -52,12 +53,17 @@ impl TemplateProgram {
     /// The string is not a valid SimplicityHL program.
     pub fn new<Str: Into<Arc<str>>>(s: Str) -> Result<Self, String> {
```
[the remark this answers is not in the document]
gerau: #208

Cargo.toml
```diff
@@ -34,6 +32,7 @@ itertools = "0.13.0"
 arbitrary = { version = "1", optional = true, features = ["derive"] }
 clap = "4.5.37"
 chumsky = "0.11.2"
+line-index = "0.1.2"
```
apoelstra: In 1ae8b0da4a3530908ee61e90e54b1fd5f2dd80f3:

This crate seems super sketchy. It uses `u32` to represent lengths all over the place, deals with this mismatch with [undocumented panics](https://docs.rs/text-size/1.1.1/src/text_size/traits.rs.html#18) (your own code uses integer type casts, which are a code smell and which occur almost nowhere else in this codebase).

It seems like this is used exclusively for a single `Display` impl? Let's just drop this dependency tree and manually count lines.

src/parse.rs
```diff
@@ -830,794 +836,973 @@ impl fmt::Display for MatchPattern {
...
+        .then(parser.recover_with(via_parser(nested_delimiters(
+            open.clone(),
+            close.clone(),
+            [
+                (Token::LParen, Token::RParen),
+                (Token::LBracket, Token::RBracket),
+                (Token::LBrace, Token::RBrace),
+                (Token::LAngle, Token::RAngle),
+            ],
+            move |_| fallback.clone(),
```
apoelstra: In 1ae8b0da4a3530908ee61e90e54b1fd5f2dd80f3:

This `clone` seems totally unnecessary; we can just move the fallback into this closure. Better, we could take the closure as an argument, and that would avoid executing the `.clone()` calls used to construct this fallback on almost every single call to `delimited_with_recovery`.

src/lexer.rs
```diff
@@ -0,0 +1,347 @@
...
+
+    let keyword = text::ident().map(|s| match s {
+        "fn" => Token::Fn,
+        "let" => Token::Let,
+        "type" => Token::Type,
+        "mod" => Token::Mod,
+        "const" => Token::Const,
+        "match" => Token::Match,
+        "true" => Token::Bool(true),
+        "false" => Token::Bool(false),
```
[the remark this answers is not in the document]
gerau: > Perhaps we should add a few reserved words here: `jet`, `witness`, `param`, and some types like `Option`.
> 
> In both these cases the type or module name is reused as a variable name, which might not be a lexer issue, but need to be checked higher up.

I'm not sure if this is an issue. Correct me if I'm wrong, but I don't think there will be a conflict on the parser side -- we can't use a variable as a type, and module names are distinct from variables because they are followed by `::`. We should, however, reserve this names for modules -- so program should not compile if there are `jet`, `witness` or `param` as the name of the module, but it should be checked higher up. 

In future it would be nice to add diagnostics about reusing this names, because it certainly not a good practice. 

> one that is almost certainly missing is `return`

Yeah, we are missing also `if` and `else` keywords.

PR #187, opened 2025-12-24
analyze_named_module: make missing modules equivalent to empty ones

### LINKED ISSUE #186: Make param and witness modules optional in SimplicityHL programs
### Project

core | compiler

## Problem

It seems to me an unnecessary boilerplate for programs without parameters or witnesses to require:
```rust
mod param {} 
mod witness {}
fn main() {}
```
## Solution
Modify `src/ast.rs` - `analyze_named_module()` function:

**Before:**
```rust
let witness_module = iter
    .next()
    .ok_or(Error::ModuleRequired(name.shallow_clone()))
    .with_span(from)?;
```

**After:**
```rust
let witness_module = iter.next();
if iter.next().is_some() {
    return Err(Error::ModuleRedefined(name)).with_span(from);
}
let mut map = HashMap::new();
if let Some(witness_module) = witness_module {
    for assignment in witness_module.assignments() {
        if map.contains_key(assignment.name()) {
            return Err(Error::WitnessReassigned(assignment.name().shallow_clone()))
                .with_span(assignment);
        }
        map.insert(
            assignment.name().shallow_clone(),
            assignment.value().clone(),
        );
    }
}
Ok(map)
```

**Why:** Stores the result without immediate error, checks for duplicates, then optionally processes the module if it exists.

**Result:** Modules are now optional instead of required.

**Labels:** `enhancement`, `usability`

apoelstra:
The web IDE expects `witness` and `params` modules, which is annoying for contracts where these modules would be empty. Allow them to be omitted, in which case they are treated as empty.

Fixes #186

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -1442,10 +1442,9 @@ fn analyze_named_module(
         ModuleItem::Module(module) if module.name == name => Some(module),
         _ => None,
     });
-    let witness_module = iter
-        .next()
-        .ok_or(Error::ModuleRequired(name.shallow_clone()))
-        .with_span(from)?;
+    let Some(witness_module) = iter.next() else {
+        return Ok(HashMap::new()); // "not present" is equivalent to empty
+    };
     if iter.next().is_some() {
         return Err(Error::ModuleRedefined(name)).with_span(from);
     }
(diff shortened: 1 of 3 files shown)


apoelstra (commented):
On f32dc20d608287c56256f04cd5b8c69779c076af successfully ran local tests

Issue #197, opened 2026-01-18
Log trace displays incorrect values for arithmetic jets


topologoanatom:
### Project version

0.4.0

### Project

tracker

### What happened?

A few jets for arithmetic operations, such as `add`, `subtract`, and `multiply`, emit incorrect values when the program tracks a **full log trace**.

## Example

```rust
fn main() {
    let x: u32 = 5;
    let y: u32 = 4;

    let sum: (bool, u32) = jet::add_32(x, y);
    let prod: u64 = jet::multiply_32(x, y);

    assert!(jet::eq_64(prod, 20));
}

```

This script emits the following results:

```text
Add32(5, 4) = (false, 18)
Multiply32(5, 4) = 40
Eq64(20, 20) = true
Verify(true) = ()

```

The problem is caused by a branch in the `tracker::parse_jet_result` function:

```rust
// Skip the leading bit when the frame has extra padding.
// This occurs because jets are wrapped in AssertL (a Case combinator),
// which adds structure to the output frame for some jets.
if output_frame.len() > target_ty.bit_width() {
    let _ = output_frame.next();
}

```

Commenting out this branch fixes the logs for arithmetic jets but breaks logs for jets like `eq_32`.

Furthermore, `output_frame.len()` is **suspiciously large** for arithmetic jets (~1000 bits). Given this, the aforementioned branch should use a more reliable way to distinguish jets that are wrapped in `AssertL`.

### Minimal reproduction steps

On this fork, https://github.com/topologoanatom/SimplicityHL/tree/bug/jet-trace, run:
```bash
cargo test test_jet_trace_regression
```
Alternatively, run any SimplicityHL script that involves add, subtract, or multiply jets with `TrackerLogLevel::Trace` and take a look at the jet result.

topologoanatom:
Changing the branch condition as follows:
```rust
// Skip the leading bit when the frame has extra padding.
// This occurs because jets are wrapped in AssertL (a Case combinator),
// which adds structure to the output frame for some jets.
if target_ty.as_sum().is_some() {
    let _ = output_frame.next();
}
```
resolves both issues, but I am not sure whether this solution is robust enough.

PR #198, opened 2026-01-20
Bug/jet trace

topologoanatom:
Fixes: #197 

### Description

Changing the heuristics for removing debug wrappers inside the `jet` result parser.

This approach is fragile, but it will suffice until the assertl-based debugging strategy is replaced.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -676,7 +676,7 @@ dependencies = [
 
 [[package]]
 name = "simplicityhl"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "arbitrary",
  "base64 0.21.3",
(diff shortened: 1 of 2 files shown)


KyrylR:
Pls follow the: https://github.com/BlockstreamResearch/SimplicityHL/blob/master/CONTRIBUTING.md

KyrylR:
@topologoanatom Commits should be signed

apoelstra:
I would also prefer that the PR description describe what's going on -- that we are changing our heuristic for removing debug wrappers in our trace logic, that this is not a great solution but in #197 we agreed it was a good thing to do for now, and that there is a version bump in this PR.

apoelstra:
Version bumps should also have a CHANGELOG entry. It looks like this release included both this PR and #187 (as well as some admin stuff that doesn't need to be in the changelog).

KyrylR:
Updated CHANGELOG https://github.com/BlockstreamResearch/SimplicityHL/pull/215

Issue #196, opened 2026-01-13
Make simc able to generate arguments and witness generation code


schoen:
### Project

compiler

### Describe the feature

Per a suggestion from @KyrylR, the SimplicityHL compiler should be able to generate Rust and/or other language code like the argument and witness generators in, for example

https://github.com/BlockstreamResearch/simplicity-contracts/tree/main/crates/contracts/src/options

This is likely to be inferrable from type annotations on witness fields in the SimplicityHL program, although it would be much easier (and more powerful) in SimplicityNG. I think the basic idea is that we enumerate all standard-looking references to saved state and witness fields in the SimplicityHL program, and then emit boilerplate code that could create the corresponding data structures based on how those references are annotated or used inside the SimplicityHL. @KyrylR can correct or clarify this.

This is relevant to wallet and app integration, and, in some settings, to obviating the need to make `.wit` files on disk.

Issue #157, opened 2025-10-29
Emitting Wallet-Friendly ABI Metadata from SimplicityHL


KyrylR:
### Project

compiler

### Describe the feature

In a nutshell: extend the SimplicityHL compiler, or build an additional analyzer, to generate a file with enough information for wallets to spend transactions locked behind a Simplicity script.

The core idea is to emit metadata during SimplicityHL compilation in the form of a JSON file (analogous to the ABI in Ethereum).

This JSON file will contain all the information needed for the wallet to spend inputs and construct outputs. Instead of writing all the code here: https://github.com/BlockstreamResearch/simplicity-contracts/blob/main/crates/cli/src/commands/options.rs, this will be handled by the wallet.

The feature is intended to be implemented in milestones.

As a first milestone, I propose the following:

Compiling the program below:
```rust
fn main() {
    jet::bip_0340_verify((param::PUBLIC_KEY, jet::sig_all_hash()), witness::SIGNATURE)
}
```

We will get a JSON file with the following fields:

```json
{
  "parameters": ["PUBLIC_KEY"],
  "witness_args": ["SIGNATURE"],
  "bip_0340_message": {
    "related_witness_arg": "SIGNATURE",
    "message": "SIG_ALL_HASH"
  },
  "script": "<script in text format>"
}
```

This will provide enough information for the wallet to verify whether it can spend the UTXO, as well as the ability to spend the UTXO overall.

The schema of the generated ABI file will depend entirely on jets.

The ABI will become more complete as more jets are supported.

schoen:
Or maybe we should move that proposal here, because this is an earlier issue? I didn't know about this one!

Is there a substantive difference in how these two issues are describing the goal and desired functionality?

PR #201, opened 2026-01-26
Add `AbiMeta` struct for `CompiledProgram` and `TemplateProgram`

ikripaka:
This pr adds possibility to generate `AbiMeta` struct both for `CompiledProgram` and `TemplateProgram`.
This struct is required for generating proper macros that will simplify creation of contracts.
Since retrieving Witness arguments and Parameter types is impossible for `CompiledProgram`, I added `AbiMeta` to improve usability for both compiled and template programs.

How it can be used in macros?

From name -> type map we can automatically generate a structs for Witness and Parameters section of contract.
This pattern is consistent across all contracts we have created so far.
https://github.com/BlockstreamResearch/SimplicityHL
By now it is in development, but I don't think that such functionality would negatively impact the code. 

Related to these issues https://github.com/BlockstreamResearch/SimplicityHL/issues/196, https://github.com/BlockstreamResearch/SimplicityHL/issues/157.

PS 
FYI, it is worth noting that functionality to retrieve the AstTreePath for program witnesses could also be added here to derive the proper selection of UTXOs.
https://github.com/ikripaka/SimplicityHL/tree/feature/wallet-abi-meta
However, this PR only adds a subset of those parameters.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,6 +95,14 @@ impl TemplateProgram {
             debug_symbols: self.simfony.debug_symbols(self.file.as_ref()),
             simplicity: commit,
             witness_types: self.simfony.witness_types().shallow_clone(),
+            parameter_types: self.simfony.parameters().shallow_clone(),
+        })
+    }
+
+    pub fn generate_abi_meta(&self) -> Result<AbiMeta, String> {
+        Ok(AbiMeta {
+            witness_types: self.simfony.witness_types().shallow_clone(),
+            param_types: self.parameters().shallow_clone(),
         })
     }
 }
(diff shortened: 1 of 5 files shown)


ikripaka:
> In [36166e7](https://github.com/BlockstreamResearch/SimplicityHL/commit/36166e7b74af027e6b7ea0c4e652d0240ac8b89a):
> 
> Definite concept NACK using `ops::Deref` for something like this.
> 
> We can add an explicit `AsRef<str>` impl if you want.

Rework it, you can see it here b82c5a35b13bf2c967b257db1d28b45177c34187

Issue #181, opened 2025-12-09
We should have an `if`/`else` construct


apoelstra:
When we're matching on a single bit it would be nicer to write this as an `if` chain.

uncomputable:
There is a Boolean match expression. You could have `if () {} else {}` as sugar for `match () { true => {}, false => {}}`.

Or do you mean an if statement without else branch? In that case, there would need to be a default Simplicity expression for the else branch.

PR #213, opened 2026-02-09
Add `if-then-else` statement support

stringhandler:
Built on #185, so that would need to be merged first.

Addresses #181


uncomputable:
This is an if-then-else expression, by the way.

stringhandler:
Converting back to draft for now. I just realised that the `Value::from_const_expr` will not exclude `If`. Which means they can be included in witness files.

PR #214, opened 2026-02-09
vscode: fix icons and `.json` files

gerau:
There is an issue with displaying icons -- `icon-theme.json` counts as a
separate icon theme, so other icons besides `simf` and `wit` didn't
displayed.

Also, I had removed activation of LSP for `.json` files, because it
works on **all** jsons, which is obviously not desired.


KyrylR:
Did you test it locally?

PR #216, opened 2026-02-10
:sparkles: Imports with namespacing, pub/use

### LINKED ISSUE #155: Add imports with namespacing, pub/use, and C3 linearization
### Project

compiler

## Reasoning

The main reasoning is to enable an analogue of the OpenZeppelin library to exist in the SimplicityHL ecosystem.

The idea is that the code will be distributed via crates.io and will allow developers to speed up the development of applications.

Primarily, it will significantly increase code reusability. If a project is heavily focused on security, it will be able to “flatten” the code and use all the functions without external dependencies.

## Semantics

- **Visibility**:
  - `pub fn` makes a function importable from other files; `fn` is file-private.
  - `pub use` re-exports imported symbols; `use` only brings them into the local scope.
  - `pub type` exports a type alias (optional but included for consistency).
- **use forms**:
  - `use m::name;`, `use m::name as alias;`, `use m::{a, b as c};`, `use m::*;`
  - `pub use m::{a, b as c};`, `pub use m::*;` (re-export)
- **Merging with C3 linearization** when resolving `pub use ...::*` across a graph of re-exports. First wins according to the MRO; conflicts require explicit aliasing.
- **Conflicts**:
  - A local `use *` that causes a collision → error unless explicitly aliased.
  - Re-export collisions are resolved by C3 order; later entries are ignored unless aliased.
- **Path resolution**: relative to the importing file’s directory, then search `-I` directories in CLI order; append `.simf` if missing.
- **Running**: Only files defining a zero-arg unit `fn main` are runnable. Libraries (no `main`) are accepted under `--check`; attempting to run them results in an error.

Overall, this feature can be delivered in milestones, so we can gradually add support.

Even though the StarkWare team introduced macros, in the long term it is much more beneficial to have official compiler support for imports. It can already be seen that their compiler version is outdated.

LesterEvSe:
## Motivation

The main goal of this PR is to enable an analogue of the OpenZeppelin library to exist in the SimplicityHL ecosystem. By introducing official compiler support for imports, we significantly increase code reusability and speed up application development. For projects heavily focused on security, this also allows developers to eventually "flatten" the code and audit all functions without external dependencies.

## Checklist

- [x] **Visibility:** Added `pub` and `pub use` to control local scope versus re-exporting.
- [x] **Import Syntax:** Implemented parsing and resolution for `use m::name;` and explicit aliasing (`as`).
- [x] **Collision Handling:** Enforced strict errors for local scope collisions, utilizing C3 linearization for resolving re-export conflicts.
- [x] **Transitive Dependencies**: Add support for libraries importing other libraries.
- [x] **Documentation:** Сode comments are updated accordingly.
- [x] **Quality Assurance:** Code is formatted, all tests pass, and CI shows no errors.


apoelstra:
Also, I would recommend splitting this up into multiple PRs.

In 3016ca28889c7767c604cc7fbd309dd78998d7d1:

This has unrelated clippy changes. But otherwise it's unobjectionable since it just adds some keywords that we definitely want to use.

In 1310a571f922aee4eebcd2e675807295c6509116:

This has a ton of lockfile noise that is unrelated to the new `tempfile` dependency.

Also, concept NACK adding `tempfile`. I've had a ton of trouble with that dependency in the past. If you want to use a dependency for this please use `xshell`, or alternately just roll your own (which I don't really recommend, but it'd be better than using `tempfile`).

I would suggest adding the new tempdir testing infrastructure in a separate commit from the actual changes. (These sorts of test infra changes can also go in their own "prepare for imports" PR.)

LesterEvSe:
Opened PR https://github.com/BlockstreamResearch/SimplicityHL/pull/264 and merged it instead.

PR #219, opened 2026-02-26
Add TemplateProgram::witness_types

egasimus:
Given nothing else but a piece of SimplicityHL source code, exposing this method allows both parameter and witness names/types to be inspected by callers of `TemplateProgram`, i.e. even before it's bound into a `CompiledProgram`.

The utility of this is further corroborated by the fact that the witness types, as defined by the user program, are already available on the private `TemplateProgram::ast` which underlies the parsing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,6 +71,11 @@ impl TemplateProgram {
         self.simfony.parameters()
     }
 
+    /// Access the witness types of the program.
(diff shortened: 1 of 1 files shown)


delta1:
reACK f87f26e70d490e3ddc01acf129dc625dc1fe762f; tested locally, provides an accessor to the witness types

PR #223, opened 2026-03-02
feat: update LSP to use new `chumsky` parser

gerau:
This PR updates LSP to use new features of `chumsky` parser. Also some functions was using `Span` struct, which was changed by parser migration, so it contains necessary changes for that too.


KyrylR (approved):
ACK 682b10ea3c1f415edc483a561c93a6425e35ca00; code review 

Note: discussed it with @gerau and decided to squash the commit to address utf16 parsing issue in the position_to_offset

PR #230, opened 2026-03-09
fix(lsp): update documentation for unwraps

gerau:
There was an error in the function signatures for `unwrap_left` and `unwrap_right` in the LSP documentation where the expected `Either` generic types were swapped.

Also make it clearer exactly which type is used when we are unwrapping these values.


KyrylR:
Rebase to main pls

Issue #204, opened 2026-01-30
Refactor errors


gerau:
As we have collection of errors in compiler, we should consider moving to the [`ariadne`](https://github.com/zesterer/ariadne) crate to pretty-print them. This also will unlock the ability to produce more complex errors that provide more context for the user.

KyrylR:
During the fix we should double check that https://github.com/BlockstreamResearch/SimplicityHL/issues/261 and https://github.com/BlockstreamResearch/SimplicityHL/pull/262 are covered as well

gerau:
In order to fully refactor the errors, I propose the following set of changes:

1. Divide the `Error` struct and its messages on a per-module basis -- parsing errors would be defined in `parse.rs`, analyzer errors would be defined in `ast.rs`, and so on. We could also integrate the `thiserror` crate for better error handling, if that is possible. This way, we would remove the `Error` enum entirely.
2. Refactor `RichError` to be more like a diagnostic struct -- it should contain fields such as: 
- A general message about the problem;
- A span;
- An ID or path to the file; 
- A vector of context, which would also contain messages and span/file locations;
- An optional note;
- A severity level.
We could also create a trait to convert the new error types from step 1 into `RichError`, so the `ErrorCollector` would do this automatically when we insert a new error. 
3. Introduce `ariadne` rendering for `RichError`. This would only require creating a method in `RichError` to build a diagnostic for the `ariadne` crate.

This, of course, would cover #261 and #262 as well.

cc @KyrylR

gerau:
Writing it here because we already have several error related issues

KyrylR:
The main thing I would separate is compiler errors from diagnostics. Per-module error types make sense, but RichError should probably become or be replaced by a diagnostic/report type that is independent of the concrete error enum. That diagnostic type can carry the message, severity, primary span, secondary context spans, notes, and source/file identity

Current RichError is doing too many jobs: domain error wrapper, source span, source content holder, chumsky parse error, and display renderer

Another nuance is that our current Span is byte-offset based while ariadne works with character offsets. So we should not pass existing spans directly into ariadne without a conversion layer, otherwise diagnostics will be wrong after non-ASCII characters (CRLF handling from #261/#262)

KyrylR:
Agree with direction, so ACK concept

gerau:
[...]

PR #262, opened 2026-03-30
fix: CRLF newline handling and error span rendering

metalurgical:
Previously, CRLF was treated as two newlines, causing incorrect line numbers, extra blank lines, and misaligned spans on Windows.

- treat CRLF as a single newline
- use consistent newline handling for line/column calculation and display
- fix trailing space rendering on empty lines
- preserve UTF-16 column alignment and span rendering
- Handle lone CR for current lexer behavior.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -5,7 +5,6 @@ use std::sync::Arc;
 use chumsky::error::Error as ChumskyError;
 use chumsky::input::ValueInput;
 use chumsky::label::LabelError;
-use chumsky::text::Char;
 use chumsky::util::MaybeRef;
 use chumsky::DefaultExpected;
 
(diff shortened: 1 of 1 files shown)


metalurgical:
> #257 was merged, could you rebase pls
> 
> Also, have to ask you to narrow down the changes to CRLF only due to the reason that the lexer is limited to the LF/CRLF/CR (\n, \r\n, \r), though I do not see a reason to support CR, as it seems a legacy thing

Has been narrowed to an explicit check instead of via `is_newline()`.

KyrylR:
Need to take a deeper look, there could be an issue with lexer that still uses `.padded()`, and Chumsky treats `\r` as whitespace there, so lone-`\r` input is still accepted

metalurgical:
> Need to take a deeper look, there could be an issue with lexer that still uses `.padded()`, and Chumsky treats `\r` as whitespace there, so lone-`\r` input is still accepted

I suspected CR still needed to be handled. So in this instance should it be erased or treated as a newline?

KyrylR:
Not sure if we ever fallback into this, but let's proceed with restoring handling of bare-\r recognition inside next_newline()

Also could you leave a short comment that this is compatibility with current lexer behavior, not a statement that CR-only files are preferred?

I believe we can add those two tests:
```rust
#[test]
fn display_with_cr_only_newlines() {
    let file = "let a: u8 = 0;\rlet b: u8 = 65536;";
    let error = Error::CannotParse("number too large to fit in target type".to_string())
        .with_span(Span::new(27, 32))
        .with_file(Arc::from(file));

    let expected = r#"
  |
2 | let b: u8 = 65536;
  |             ^^^^^ Cannot parse: number too large to fit in target type"#;

    assert_eq!(&expected[1..], &error.to_string());
}
```

```rust
#[test]
fn display_span_as_point_on_trailing_cr_only_empty_line() {
    let file = "fn main(){\r    let a:\r";
    let error = Error::CannotParse("eof".to_string())
        .with_span(Span::new(file.len(), file.len()))
        .with_file(Arc::from(file));

    let expected = r#"
  |
3 |
  | ^ Cannot parse: eof"#;

    assert_eq!(&expected[1..], &error.to_string());
}
```

metalurgical:
> Not sure if we ever fallback into this, but let's proceed with restoring handling of bare-\r recognition inside next_newline()
> 
> Also could you leave a short comment that this is compatibility with current lexer behavior, not a statement that CR-only files are preferred?



> Not sure if we ever fallback into this, but let's proceed with restoring handling of bare-\r recognition inside next_newline()
> 
> Also could you leave a short comment that this is compatibility with current lexer behavior, not a statement that CR-only files are preferred?

Lone CR support was added back in for compatibility with current lexer behavior only. 

The proposed tests were also added in.

KyrylR:
@gerau Could you take a look as well here?

gerau:
ACK [`079afeb`](https://github.com/BlockstreamResearch/SimplicityHL/pull/262/commits/079afebd900dd88f1f6fb8531b3f37e380161394)

KyrylR:
@metalurgical sorry for the delay, with recent addition of modules we learned that we will have to do the overhaul of error handling, closing this issue, and linking it in #204, so we can use it as inspiration or reopen it later

metalurgical:
No problem

PR #263, opened 2026-03-31
Versioning

### LINKED ISSUE #248: Feature: Versioning
### Project

compiler

### Describe the feature

Current version of Simplicity is 
1. Not backward compatible
2. Each user could produce different hex-string depending on the crate versions, hardware or etc.
Since that we need to add versioning to simplicity, which will help to keep determinism and keep ability not to rewrite each smart-contract whether the compiler changes.

The proposed versioning solution is to keep Solidity's `pragma` version usage. So we would write `pragma version X.Y.Z;` at the top of the file, and it will lock the compilation to those rules.

Implementation plan:
1. Update the lexer and parser to read the pragma at the start of the file.
2. Read the compiler's actual version from `Cargo.toml` at build time.
3. Show version errors if there is a mismatch (for example, if a user tries to use new functions in an old version, or old functions in a new one).
4. Create a tool that reads the pragma and automatically downloads and executes the correct compiler binary for the requested version.
5. To handle cases where a user tries to use a newer feature on an older compiler, update error messages to include a helpful hint to check their pragma and compiler version. (Reads an error message after compiler work, so that each binary is deterministic)

Sdoba16:
Closes #248 

**Add compiler versioning for SimplicityHL contracts**

This PR introduces strict compiler version requirements

* **Version Directives:** Every `.simf` file must now begin with a `simc "..."` directive. The directive is safely extracted (ignoring comments) and validated before full AST construction.
* **SemVer Ranges:** Added support for standard Semantic Versioning ranges (e.g., `>=0.6.0`, `^0.6.0`, `~0.6.0`) 
* **Dependency Versioning:** The compiler driver now verifies version compatibility across the entire multi-file project (`main.simf` and all imported `--dep` libraries)
* **Developer Tooling:** Added `update_examples.py` and a CI workflow to easily manage version across all examples and tests.

Syntax:
```rust
simc ">=0.6.0";

fn main() {
    ...
}
```

Needed simplex addition for this PR:
* **Version Installer:** Tooling that will go through the dependencies and look for best fitting version to be installed. The tool will take existing versions from HL releases and install the best-fitted one.

### DIFF
--- a/doc/versioning.md
+++ b/doc/versioning.md
@@ -1,0 +1,33 @@
+# Compiler versioning
+
+Every `.simf` file may begin with a compiler version directive:
+
+```text
+simc ">=0.6.0";
+```
+
+It is an optional fail-fast check: if the running compiler does not satisfy the
+declared [SemVer](https://semver.org) range, compilation stops with a clear message
+before lexing — even when the rest of the file is written for a newer language
+version. Without a directive the file still compiles and `simc` prints a warning.
+The directive must be the first non-comment item, at most once per file; `simc` is
+a reserved keyword. In multi-file projects the entry file and every reachable
+dependency are checked.
+
+Range versions are plain `x.y.z` numbers: pre-release tags are not allowed in a
+directive, and a pre-release compiler (e.g. `0.6.0-rc.2`) counts as its base
+version.
+
+## Reproducibility
+
+A range does not pin the output: different compiler versions can satisfy it and
+produce different Commitment Merkle Roots (CMRs), hence different addresses. Pin an
+exact version (`=x.y.z`) for anything you deploy. Selecting, pinning, and fetching
+compilers is the job of higher-level tooling, not `simc`.
+
+## Details
+
+See the rustdoc of `src/version.rs`: the checking flow, the frozen directive
+syntax, pre-release handling, and the tooling entry point
+(`SimcDirective::requirement_of`) are documented next to the code, so they cannot
+drift.
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -29,6 +29,9 @@
 mod linearization;
 mod resolve_order;
 
+#[cfg(test)]
+mod version_tests;
+
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::path::PathBuf;
 use std::sync::Arc;
--- a/src/driver/version_tests.rs
+++ b/src/driver/version_tests.rs
@@ -1,0 +1,257 @@
+//! Multi-file enforcement of `simc "...";` directives: the entry file and every
+//! reachable dependency are checked. Semver matching itself is covered by
+//! `crate::version`'s unit tests.
+
+use crate::driver::tests::setup_graph_raw;
+
+/// Builds the given files (with `{v}` replaced by the current compiler version)
+/// through the dependency-graph build, returning success and collected diagnostics.
+fn build(files: &[(&str, &str)]) -> (bool, String) {
+    // Ranges cannot name pre-releases, so an `-rc` compiler substitutes its base version.
+    let v = env!("CARGO_PKG_VERSION").split('-').next().unwrap();
+    let owned: Vec<(&str, String)> = files
+        .iter()
+        .map(|(p, c)| (*p, c.replace("{v}", v)))
+        .collect();
+    let refs: Vec<(&str, &str)> = owned.iter().map(|(p, c)| (*p, c.as_str())).collect();
+    let (graph_opt, _, _ws, handler) = setup_graph_raw(refs);
+    let ok = graph_opt.is_some() && !handler.has_errors();
+    (ok, handler.to_string())
+}
+
+fn assert_builds(files: &[(&str, &str)]) {
+    let (ok, errors) = build(files);
+    assert!(ok, "build failed unexpectedly. Errors:\n{errors}");
+}
+
+/// Like [`assert_builds`], but the build must fail with a diagnostic containing
+/// `expected_err`.
+fn assert_build_fails(expected_err: &str, files: &[(&str, &str)]) {
+    let (ok, errors) = build(files);
+    assert!(!ok, "build succeeded when it should have failed");
+    assert!(
+        errors.contains(expected_err),
+        "Expected error containing '{expected_err}' but got:\n{errors}"
+    );
+}
+
+/// A multi-file program whose every file declares a compatible directive compiles:
+/// each directive is checked and stripped, and the bodies still parse across `use`.
+#[test]
+fn mixed_valid_operators() {
+    assert_builds(&[
+        (
+            "main.simf",
+            r#"simc "^{v}";
+use lib::A::foo;
+fn main() {}"#,
+        ),
+        (
+            "libs/lib/A.simf",
+            r#"simc "={v}";
+use crate::B::foo;
+pub fn foo() {}"#,
+        ),
+        (
+            "libs/lib/B.simf",
+            r#"simc ">0.1.0";
+use crate::C::foo;
+pub fn foo() {}"#,
+        ),
+        (
+            "libs/lib/C.simf",
+            r#"simc "*";
+pub fn foo() {}"#,
+        ),
+    ]);
+}
+
+/// The entry file's directive is checked.
+#[test]
+fn main_too_old_fails() {
+    assert_build_fails(
+        "Incompatible compiler version",
+        &[
+            (
+                "main.simf",
+                r#"simc ">99.0.0";
+use lib::A::foo;
+fn main() {}"#,
+            ),
+            (
+                "libs/lib/A.simf",
+                r#"simc "={v}";
+pub fn foo() {}"#,
+            ),
+        ],
+    );
+}
+
+/// Every reachable dependency's directive is checked, not just the entry file.
+#[test]
+fn lib_too_old_fails() {
+    assert_build_fails(
+        "Incompatible compiler version",
+        &[
+            (
+                "main.simf",
+                r#"simc "={v}";
+use lib::A::foo;
+fn main() {}"#,
+            ),
+            (
+                "libs/lib/A.simf",
+                r#"simc ">99.0.0";
+pub fn foo() {}"#,
+            ),
+        ],
+    );
+}
+
+/// A file that is never imported is not checked, even with an incompatible directive.
+#[test]
+fn unreferenced_file_with_invalid_version_ignored() {
+    assert_builds(&[
+        (
+            "main.simf",
+            r#"simc "={v}";
+use lib::A::foo;
+fn main() {}"#,
+        ),
+        (
+            "libs/lib/A.simf",
+            r#"simc "={v}";
+pub fn foo() {}"#,
+        ),
+        (
+            "libs/lib/B.simf",
+            r#"simc ">99.0.0";
+pub fn unused() {}"#,
+        ),
+    ]);
+}
+
+/// An omitted directive is allowed through the driver: only a present directive is
+/// enforced, so a directive-less entry file builds successfully.
+#[test]
+fn directive_omitted() {
+    assert_builds(&[("main.simf", "fn main() {}")]);
+}
+
+/// The compatibility check runs on the raw text, before lexing: an incompatible
+/// compiler is reported even when the file's body cannot be tokenized (here a string
+/// literal, which the language does not have), instead of a pile of lex errors.
+#[test]
+fn incompatible_reported_despite_unlexable_body() {
+    assert_build_fails(
+        "Incompatible compiler version",
+        &[(
+            "main.simf",
+            r#"simc ">99.0.0";
+fn main() { let s = "future syntax"; }"#,
+        )],
+    );
+}
+
+/// A stray directive is reported alone: its `"<range>";` remnant and the rest of
+/// the parse must not add noise on top of the reserved-keyword error.
+#[test]
+fn stray_directive_reports_single_error() {
+    let (ok, errors) = build(&[(
+        "main.simf",
+        r#"simc "={v}";
+simc "={v}";
+fn main() {}"#,
+    )]);
+    assert!(!ok, "duplicate directive must fail the build");
+    assert!(
+        errors.contains("reserved"),
+        "expected the reserved-keyword error, got:\n{errors}"
+    );
+    assert!(
+        !errors.contains("Cannot parse"),
+        "remnant noise must be suppressed, got:\n{errors}"
+    );
+}
+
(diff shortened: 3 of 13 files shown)


stringhandler (commented):
Concept ACK.

I've come around to the idea that users should optionally be able to specify the compiler version. However, I don't like the `pragma` keyword as it deviates from rust, which we haven't done previously. Below are some suggestions. 

I also don't think it should be required in the `.simc`, which will allow us to specify it in a project level manifest for when multiple files are included and for libraries in future. Also, the user should not be forced to provided it even in a single file scenario. This is just extra friction for new developers.

apoelstra:
ab41feca29a8dc1c27d8a865c7f1be7319fb92d1 needs rebase

KyrylR:
Do we have an issue for discussion this?

Sdoba16:
> Do we have an issue for discussion this?

https://github.com/BlockstreamResearch/SimplicityHL/issues/248

KyrylR:
Needs rebase

apoelstra:
c56e75db436dbf41f20c613d84decf89e49882a9 needs rebase

doc/versioning.md
```diff
@@ -0,0 +1,40 @@
+# Compiler Versioning
+
+The SimplicityHL compiler enforces strict version compatibility to prevent contracts from silently breaking due to compiler updates, semantic changes, or new language features. 
+
+Every `.simf` file must begin with a compiler version directive:
+```rust
+#![compiler_version(">=0.6.0")]
```
LesterEvSe: This syntax suggests we have a preprocessing phase, which is not true. Since SimplicityHL is a Rust-like language, users will see `#![...]` and assume it works like Rust's preprocessor.

Let's avoid the confusion and switch to `simc X.Y.Z`. Since `pragma version` is not approved yet, we are looking at variants like `compiler`, `compiler-version`, or `simc`. In my opinion, `simc` is the most optimal variant because it is concise and saves keystrokes for the user.

doc/versioning.md
```diff
@@ -0,0 +1,40 @@
...
+
+## For Compiler Contributors
+
+If you are contributing to the SimplicityHL compiler and bumping the version in `Cargo.toml`, you must update the version directives in all examples and tests. 
+
+You can do this automatically from the project root:
+```bash
+python3 update_examples.py <new_version>
+```
+See `CONTRIBUTING.md` for more details.
```
LesterEvSe: Regarding the `update_examples.py` script and our test contracts, I do not see a strict need for pinned compiler versions here.

We can simplify our workflow by using a range like `pragma version >=0.5.0` across all test files. If a specific contract introduces breaking changes or becomes incompatible in the future, we can simply bump the minimum requirement for that file alone. Otherwise, we can leave them as they are and avoid running an update script every time the compiler version changes.

src/driver/version_tests.rs
```diff
@@ -0,0 +1,394 @@
+use crate::driver::tests::setup_graph_raw;
```
LesterEvSe: Let's add a quick note to the docstring explaining the motivation behind this macro

src/driver/version_tests.rs
```diff
@@ -0,0 +1,394 @@
...
+                assert!(graph_opt.is_some() && !handler.has_errors(),
+                    "Scenario failed unexpectedly. Errors:\n{}", handler);
+            } else {
+                assert!(graph_opt.is_none() || handler.has_errors(),
+                    "Scenario succeeded when it should have failed.");
+                let expected_err: Option<&str> = $expected_err;
+                if let Some(err) = expected_err {
+                    assert!(handler.to_string().contains(err),
+                        "Expected error containing '{}' but got:\n{}", err, handler);
+                }
```
LesterEvSe: The nested blocks can be rewritten as follows:

``` Rust
if $expect_success {
    assert!(graph_opt.is_some() && !handler.has_errors(),
        "Scenario failed unexpectedly. Errors:\n{}", handler);
    return;
}

assert!(graph_opt.is_none() || handler.has_errors(),
    "Scenario succeeded when it should have failed.");

let expected_err: Option<&str> = $expected_err;
if let Some(err) = expected_err {
    assert!(handler.to_string().contains(err),
        "Expected error containing '{}' but got:\n{}", err, handler);
}
```

LesterEvSe (changes requested):

stringhandler:
Is it required to add the compiler version? I would say the examples should only specify the minimum version needed to compile, which for all of these would not be necessary as they don't have any breaking changes. The only ones that might need to be specified are the ones using the `use` and `pub` keywords.

It is generally fine to include a compiler version, but these look like they need to keep being updated.

examples/local_crate/main.simf
```diff
@@ -1,3 +1,5 @@
+simc ">=0.6.0";
```
stringhandler: Should this not just be `simc "0.6.0"`?

Can we guarantee it will work on `0.7.0`?

doc/versioning.md
```diff
@@ -0,0 +1,30 @@
+# Compiler Versioning
+
+The SimplicityHL compiler enforces strict version compatibility to prevent contracts from silently breaking due to compiler updates, semantic changes, or new language features. 
+
+Every `.simf` file must begin with a compiler version directive:
```
stringhandler:
```suggestion
Every `.simf` file may begin with a compiler version directive:
```

I think it is incredibly limiting to new developers to enforce this. Surely this should be opt-in.

src/driver/mod.rs
```diff
@@ -356,11 +363,18 @@ pub(crate) mod tests {
         // Create all requested files
         for (path, content) in files {
             let full_path = format!("workspace/{}", path);
-            let created_file = canon(&ws.create_file(&full_path, content));
+            let injected_content = if crate::version::has_version_header(content)
```
stringhandler: This could use a comment.

src/driver/version_tests.rs
```diff
@@ -0,0 +1,402 @@
...
+                assert!(handler.to_string().contains(err),
+                    "Expected error containing '{}' but got:\n{}", err, handler);
+            }
+        }
+    };
+}
+
+version_test!(
+    exact_match_all, true, None,
+    "main.simf" => "simc \"={v}\";\nuse lib::A::foo;\nfn main() {}",
```
stringhandler:
```suggestion
    "main.simf" => r#"simc ""={v}"";
    use lib::A::foo;
    fn main() {}"#,
```

These would be much easier to review if you use raw strings

apoelstra:
cd5b1e96ad60102e538fd9d39310c3d8601b2066 needs rebase

examples/local_crate/main.simf
LesterEvSe: I explained the reason here: https://github.com/BlockstreamResearch/SimplicityHL/pull/263#discussion_r3354876757

doc/versioning.md
LesterEvSe: I do not fully understand how it will work. If we load a file without a version, then how are we supposed to determine it? Do we assume that it is always correct? What if the functionality in that file becomes deprecated? What if we have some vulnerabilities in the file? Thus, I think we need to make it mandatory for users to use versions

stringhandler: Let me walk through the two concrete scenarios:

Case 1: File uses newer syntax (e.g., pub or use, introduced in simc 0.6.0)
If no version is specified, the file simply fails to compile on older compilers with a syntax error. The simc field doesn't change this outcome - it just gives a much better error message ("this file requires simc >= 0.6.0") instead of a confusing parser error.

Case 2: File uses deprecated or removed functionality
Same story: without a version constraint, the file fails to compile once that feature is removed. The simc field lets authors signal simc >=0.5.0, <0.6.0 to make that explicit, but it's still an error either way.

On vulnerabilities: the simc version field controls compilation compatibility, not runtime security. Vulnerabilities in Simplicity programs are addressed at the consensus/jet layer, independently of the compiler version.

So omitting the simc field is never silently wrong - the compiler still catches incompatibilities. Making it mandatory adds friction for new developers who just want to write a simple program, and requires them to understand version semantics before writing their first line of code. Keeping it as may means it's there when you need better error messages, but doesn't block the happy path.

LesterEvSe: Got it, okay

src/version.rs
```diff
@@ -0,0 +1,482 @@
...
+                pre: semver::Prerelease::EMPTY,
+                ..version.clone()
+            }
+        }
+    }
+}
+
+/// Replace the directive with equal-length spaces so the parser never sees it
+/// while later byte offsets (and thus error spans) stay correct.
+pub(crate) fn blank_version_directive(content: &str) -> Cow<'_, str> {
```
LesterEvSe: This function answers the question "What does this function do?", but not "Why was this function needed?". I looked at the following commits to see its usage inside the parse.rs file, but I still do not understand the reason for using it

src/version.rs
```diff
@@ -0,0 +1,482 @@
...
+    /// A single well-formed directive, correctly placed before any program code.
+    Found { req: &'a str, span: Span },
+    /// No directive anywhere in the file.
+    Absent,
+    /// A directive-looking line that is not a well-formed `simc "...";`.
+    Malformed(Span),
+    /// More than one directive.
+    Duplicate(Span),
+    /// A well-formed directive that appears after program code.
+    Misplaced(Span),
```
LesterEvSe: Let's stick to the style that was mentioned here: https://github.com/BlockstreamResearch/SimplicityHL/pull/328#issuecomment-4534785428

LesterEvSe (changes requested):

src/tracker.rs
```diff
@@ -451,8 +451,9 @@ mod tests {
 
     #[test]
     fn test_debug_and_jet_tracing() {
+        let program_text = TEST_PROGRAM;
```
LesterEvSe: I do not think we need to change this file at all

src/version.rs
```diff
@@ -0,0 +1,404 @@
...
+                }
+                (_, Some(close)) => {
+                    depth -= 1;
+                    rest = &rest[close + "*/".len()..];
+                }
+                _ => return "",
+            }
+        }
+        rest
+    }
```
LesterEvSe: These `skip_trivia` and `skip_block_comment` helpers re-implement the comment handling already in the `lexer.rs:208-223`.

Since both must agree on comment syntax, could we extract that logic into one shared trivia recognizer and call it from both places?

src/lib.rs
```diff
@@ -1530,6 +1531,47 @@ fn main() {
...
+    /// present, so a directive-less program compiles. The dependency-path equivalent is
+    /// covered end to end by the CLI test `cli_version_missing_warns_but_compiles`.
+    #[test]
+    fn directive_omitted() {
+        let result = TemplateProgram::new(
+            "fn main() {}",
+            Box::new(crate::ast::ElementsJetHinter::new()),
+        );
+        assert!(result.is_ok(), "expected success, got: {:?}", result.err());
+    }
```
LesterEvSe: Let's delete it, because we have the same test at `version_tests.rs:136`

```suggestion
```

LesterEvSe:
@apoelstra, we are adding `semver` to the `Cargo.toml` file. What do you think about it? Maybe you have experience with it and know more about which one to use? We would be glad to hear your thoughts on it

apoelstra:
I think it's fine. `semver` lives in David Tolnay's personal Github account, but it's used by `cargo` and extremely widely deployed. But [it does have some super weird behavior around alpha/beta/rc suffixes](https://github.com/rust-bitcoin/rust-bitcoin/issues/5358#issuecomment-3703064717) (see that comment and the two after it).

So I would strongly suggest that we enforce that any semver versions we accept contain only numbers and `.`s, which should keep us within the "normal" behavior of the `semver` crate.

Using `semver` will also cause us to inherit Rust's "In x.y.z, x is the major version, except in 0.y.z, y is the major version" behavior, which I think is fine (good, even). Though FYI this is not part of the official semver standard. I believe that there is some further weirdness for 0.0.z, but anybody using those versions for real code is asking for trouble, so that's fine.

src/parse.rs
```diff
@@ -1243,14 +1248,36 @@ impl<A: ChumskyParse + crate::unstable::RequireFeature + std::fmt::Debug> ParseF
...
+        // A stray `simc` makes every other diagnostic noise — its `"<range>";` remnant
+        // does not lex — so the reserved-keyword errors are reported alone.
+        if lex_errs
+            .iter()
+            .any(|e| matches!(e.error(), Error::ReservedSimcKeyword))
+        {
+            lex_errs.retain(|e| matches!(e.error(), Error::ReservedSimcKeyword));
+            handler.extend(source, lex_errs);
+            return None;
+        }
```
LesterEvSe, line 1281: Let's make the same check for the `parse_from_str` function.

tests/cli.rs
```diff
@@ -87,3 +111,189 @@ fn cli_reserved_crate_mapping_fails() {
         stderr
     );
 }
+
+/// The output carries the version of the compiler that produced it, so the exact
+/// compiler can be identified from a shared artifact (different versions can
+/// produce different CMRs from the same source).
+#[test]
```
LesterEvSe, line 119: Since this test uses `.arg("--json")`, we assume that the `serde` crate is enabled. But a plain `#[test]` does not enable this feature, so the tests fail on the CI matrix. So let's fix it:

```suggestion
#[cfg(feature = "serde")]
#[test]
```

Sdoba16:
> I think it's fine. `semver` lives in David Tolnay's personal Github account, but it's used by `cargo` and extremely widely deployed. But [it does have some super weird behavior around alpha/beta/rc suffixes](https://github.com/rust-bitcoin/rust-bitcoin/issues/5358#issuecomment-3703064717) (see that comment and the two after it).
> 
> So I would strongly suggest that we enforce that any semver versions we accept contain only numbers and `.`s, which should keep us within the "normal" behavior of the `semver` crate.
> 
> Using `semver` will also cause us to inherit Rust's "In x.y.z, x is the major version, except in 0.y.z, y is the major version" behavior, which I think is fine (good, even). Though FYI this is not part of the official semver standard. I believe that there is some further weirdness for 0.0.z, but anybody using those versions for real code is asking for trouble, so that's fine.

Now we support only numbers and `.s`. For pre-releases, we strip the compiler's `-rc` tag before matching, so an `x.y.z` pre-release still satisfies `>=x.y.z`.

LesterEvSe (approved):
ACK 64086e3a1ccafc6a1d918dca141f0b4f89e04170; tested locally with `just check` and `just check_fuzz`

KyrylR (approved):
ACK 42ffbd234f007e7fda84676da36a18542b461fc4; successfully ran local tests

Issue #270, opened 2026-04-03
Refactor(ErrorHandler): Consolidate generic `String` errors into structured type


LesterEvSe:
Currently, several functions across return a generic `String` as an error (e.g., `Result<T, String>`). While sufficient for early development, this is too generic for robust compiler error handling. It prevents us from easily matching on specific error variants or attaching detailed span context for diagnostics.

https://github.com/BlockstreamResearch/SimplicityHL/pull/269#discussion_r3032520512

PR #271, opened 2026-04-03
Implement `DependencyGraph` and foundational driver architecture

LesterEvSe:
**Dependency Graph:** Added `driver.rs` featuring the `DependencyGraph` implementation to build a Directed Acyclic Graph of project dependencies.
**Trait Refactoring:** Updated the `ParseFromStrWithErrors` trait to accept a full `SourceFile` (containing both metadata and source text) instead of just the raw string content.

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -1,0 +1,358 @@
+use std::collections::{HashMap, HashSet, VecDeque};
+use std::path::PathBuf;
+use std::sync::Arc;
+
+use chumsky::container::Container;
+
+use crate::error::{Error, ErrorCollector, RichError, Span};
+use crate::parse::{self, ParseFromStrWithErrors};
+use crate::resolution::{CanonPath, DependencyMap, SourceFile};
+
+/// Represents a single, isolated file in the SimplicityHL project.
+/// In this architecture, a file and a module are the exact same thing.
+#[derive(Debug, Clone)]
+struct Module {
+    source: SourceFile,
+    /// The completely parsed program for this specific file.
+    /// it contains all the functions, aliases, and imports defined inside the file.
+    parsed_program: parse::Program,
+}
+
+/// An Intermediate Representation that helps transform isolated files into a global program.
+///
+/// While an AST only understands a single file, the `DependencyGraph` links multiple
+/// ASTs together into a Directed Acyclic Graph (DAG). This DAG is then used to build
+/// one convenient `Program` struct for the semantic analyzer can easily process.
+///
+/// This structure provides the global context necessary to solve high-level compiler
+/// problems, including:
+/// * **Cross-Module Resolution:** Allowing the compiler to traverse edges and verify
+///   that imported symbols, functions, and types actually exist in other files.
+/// * **Topological Sorting:** Guaranteeing that modules are analyzed and compiled in
+///   the strictly correct mathematical order (e.g., analyzing module `B` before module
+///   `A` if `A` depends on `B`).
+/// * **Cycle Detection:** Preventing infinite compiler loops by ensuring no circular
+///   imports exist before heavy semantic processing begins.
+pub struct DependencyGraph {
+    /// Implements the Arena Pattern to act as the sole, centralized owner of all parsed modules.
+    ///
+    /// In C++ or Java, a graph would typically link dependencies using direct memory
+    /// pointers (e.g., `List<Module*>`). In Rust, doing this requires either
+    /// lifetimes or performance-heavy reference counting (`Rc<RefCell<T>>`).
+    ///
+    /// Using a flat `Vec` as a memory arena is the idiomatic Rust solution.
+    modules: Vec<Module>,
+
+    /// The configuration environment.
+    /// Used to resolve external library dependencies and invoke their associated functions.
+    dependency_map: Arc<DependencyMap>,
+
+    /// Fast lookup: `CanonPath` -> Module ID.
+    /// A reverse index mapping absolute file paths to their internal IDs.
+    /// This solves the duplication problem, ensuring each file is only parsed once.
+    lookup: HashMap<CanonPath, usize>,
+
+    /// Fast lookup: Module ID -> `CanonPath`.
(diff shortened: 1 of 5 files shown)


src/driver.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// The completely parsed program for this specific file.
+    /// it contains all the functions, aliases, and imports defined inside the file.
+    pub parsed_program: parse::Program,
+}
+
+/// The Dependency Graph itself.
+pub struct ProjectGraph {
+    /// Arena Pattern: the data itself lives here.
+    /// A flat vector guarantees that module data is stored contiguously in memory.
+    #[expect(dead_code)]
```
metalurgical:
```suggestion
    #[allow(dead_code)]
```

src/driver/mod.rs
```diff
@@ -0,0 +1,327 @@
+use std::collections::{HashMap, VecDeque};
```
KyrylR: Could we move it to driver/mod.rs? 
The reasoning is that we gonna introduce a few independent algorithms here, so we could have them under same module, as they are just related to the driver itself

src/driver.rs
```diff
@@ -0,0 +1,327 @@
...
+    ///
+    /// * `Ok(Some(Self))` - If the entire project graph was successfully resolved and parsed.
+    /// * `Ok(None)` - If the graph traversal completed, but one or more modules contained
+    ///   errors (which have been safely logged into the `handler`).
+    ///
+    /// # Errors
+    ///
+    /// This function will return an `Err(String)` only for critical internal compiler errors
+    /// (e.g., if a provided `SourceFile` is unexpectedly missing its underlying file path).
+    pub fn new(
```
KyrylR, line 93: Let's have pub functions first then private

src/driver.rs
```diff
@@ -0,0 +1,327 @@
...
+        let root_id = 0;
+        graph.lookup.insert(root_name, root_id);
+        graph.dependencies.insert(root_id, Vec::new());
+
+        let mut queue = VecDeque::new();
+        queue.push_back(root_id);
+
+        while let Some(curr_id) = queue.pop_front() {
+            // We need this to report errors inside THIS file.
+            let importer_source = graph.modules[curr_id].source.clone();
```
KyrylR: Let's handle modules[curr_id], via let statement to avoid panic

src/driver/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+            graph.load_and_parse_dependencies(
+                curr_id,
+                valid_imports,
+                &importer_source,
+                handler,
+                &mut queue,
+            );
+        }
+
+        Ok((!handler.has_errors()).then_some(graph))
```
KyrylR, line 171: Is it expected that `ParseFromStrWithErrors` now returns `None` whenever the shared `ErrorCollector` already contains any earlier error, and `DependencyGraph::new` reuses one collector across the entire BFS?

src/driver/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+        importer_source: &SourceFile,
+        handler: &mut ErrorCollector,
+        queue: &mut std::collections::VecDeque<usize>,
+    ) {
+        for (path, import_span) in valid_imports {
+            if let Some(&existing_id) = self.lookup.get(&path) {
+                let deps = self.dependencies.entry(curr_id).or_default();
+                if !deps.contains(&existing_id) {
+                    deps.push(existing_id);
+                }
```
KyrylR, line 253: Does it actually detect import cycles? As it is written it treats any already-seen path as a normal resolved dependency and continues. For `A -> B -> A`, B’s import of A hits the `existing_id` branch, and in `Ok((!handler.has_errors()).then_some(graph))` still returns Some(graph) with no diagnostic

LesterEvSe: No, it will be the responsibility of C3 and the other algorithms in the following PR

src/driver/mod.rs
LesterEvSe: This was not expected. I could add a local `ErrorCollector` inside the `parse_and_get_program` function to parse more files and collect more errors. I think this would be better than the current approach

KyrylR: Yes, then you could return Result<Self, ErrorCollector>

src/driver/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+                continue;
+            }
+
+            let Some(module) =
+                Self::parse_and_get_program(&path, importer_source.clone(), import_span, handler)
+            else {
+                continue;
+            };
```
KyrylR, line 262: If parse_and_get_program fails here, the path is never recorded in lookup, so another parent importing the same file will try to parse it again and emit the same lex or syntax errors again

We could try to fix it here, or have a GitHub issue for it

src/driver/mod.rs
LesterEvSe: That won't work with the collector because the Strings are not bound to the `span` elements that the collector requires. I took a slightly different approach and created the collector in the `parse_and_get_program` function to collect more errors.

I think it will be fixed once the #270 issue has been resolved.

src/driver/mod.rs
LesterEvSe: Tried to resolve it here

src/driver/mod.rs
KyrylR: Add a TODO here, and mention in the comment the issue

KyrylR (approved):
ACK ae3ed687349d04f0be7bbd0a2d9a02f4ab98d819; code review, comments

apoelstra:
ae3ed687349d04f0be7bbd0a2d9a02f4ab98d819 needs rebase

KyrylR:
> https://github.com/BlockstreamResearch/SimplicityHL/commit/ae3ed687349d04f0be7bbd0a2d9a02f4ab98d819 needs rebase

Resolved locally

LesterEvSe:
I have reopened the PR for fixes after all. #278

PR #272, opened 2026-04-03
fix: enable fuzzing with cargo-fuzz

metalurgical:
This enables fuzzing of targets using cargo-fuzz on github actions where it previously was ignored.

Changes:
- Install cargo fuzz
- Run cargo fuzz for each target

Can be updated to use mapfile or parameterized in future.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -189,3 +189,14 @@ jobs:
         run: |
           cargo check --manifest-path=fuzz/Cargo.toml
 
+      - name: "Install cargo-fuzz"
+        run: |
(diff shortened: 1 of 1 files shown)


metalurgical:
resolves #267 
closes #267

apoelstra (approved):
ACK f7af079880417770242d9f3612d2cf333bf722c9; successfully ran local tests; simple and straightforward

KyrylR (approved):
ACK https://github.com/BlockstreamResearch/SimplicityHL/commit/f7af079880417770242d9f3612d2cf333bf722c9; successfully ran local tests; code review

Issue #250, opened 2026-03-20
Allow more than two arms in match statement


schoen:
### Project

core

### Describe the feature

This is a somewhat challenging or at least somewhat complex feature, and I wouldn't assume it could be completed right away.

Witnesses very often use nested `Either` types to let a transaction creator select one of many code paths inside of the SimplicityHL contract. But right now, we have to use nested `match` expressions as a result. An example is

https://github.com/BlockstreamResearch/simplicity-contracts/blob/e6d97ad5e5f3933005027b8d127cdb314d16fd78/crates/contracts/src/finance/options/source_simf/options.simf#L381

where we have four actions that can be taken and as a result a nested `match`.

It would be great to be able to have one top-level `match` that could handle everything, like

```
match witness::thingy {
    Left(Left(x: u8)) => one_thing(x),
    Left(Right(x: u8)) => second_thing(x),
    Right(Left((x, y): (u8, u256)) => third_thing(x, y),
    Right(Right(s: signature)) => fourth_thing(signature),
}
```

I think this would help make things a lot more readable in these cases. @stringhandler also has an idea for `Enum`s which would also make things more readable in this situation (and would also require more than two `match` arms to be supported).

KyrylR:
@schoen maybe #376 will be able to close this issue

PR #275, opened 2026-04-06
feat: implement structured pattern matching with nested constructor support

metalurgical:
Introduce a full pattern matching pipeline for match expressions, including typed pattern representation, validation, and a structured decision tree.

This establishes a foundation for deterministic and type-safe match evaluation, enabling correct handling of nested patterns and complex destructuring.

Additionally extract tests for parse to new file to reduce file bloat and improve maintainability.

### Overview Diagram

```mermaid
flowchart TD
    A["match syntax"] --> B{"Old vs New"}

    B -->|Old| C["2 arms only → normalize"]
    B -->|New| D["multiple arms → analyze → validate → compile"]

    C --> E["Match.left/right"]
    D --> E
```

---

### Expanded Diagram

```mermaid
flowchart TD
    A["match syntax"] --> B{"Old vs New"}

    B -->|Old| C1["exactly 2 arms required"]
    C1 --> C2["classify arm pair"]
    C2 --> C3["Left/Right or None/Some or False/True"]
    C3 --> E["Match.left/right"]

    B -->|New| D1["parse multiple arms"]
    D1 --> D2["convert to structured pattern representation"]
    D2 --> D3["assign types to patterns"]

    D3 --> D4["ensure all arms use the same match kind"]
    D4 --> D5["check for duplicate arms"]
    D5 --> D6["check for overlapping patterns"]
    D6 --> D7["check for unreachable arms"]
    D7 --> D8["check exhaustiveness"]

    D8 --> D9["build decision tree"]
    D9 --> D10["convert back to two-arm form"]
    D10 --> E["Match.left/right"]
```

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -406,6 +406,142 @@ impl Match {
 
 impl_eq_hash!(Match; scrutinee, left, right);
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+enum ParsedMatchCtor {
+    Left,
+    Right,
+    Some,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+enum ParsedMatchPattern {
+    False,
+    True,
+    None,
+    Ctor(ParsedMatchCtor, Box<ParsedMatchPattern>),
+    Bind(Pattern, AliasedType),
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+struct ParsedMatchArm {
+    pattern: ParsedMatchPattern,
+    expression: Arc<Expression>,
+    span: Span,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
+enum MatchFamily {
+    Either,
+    Option,
+    Bool,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
+enum PatternCtorTag {
+    Left,
+    Right,
+    Some,
+    None,
+    False,
+    True,
+}
+
+impl PatternCtorTag {
+    fn family(self) -> MatchFamily {
+        match self {
+            PatternCtorTag::Left | PatternCtorTag::Right => MatchFamily::Either,
+            PatternCtorTag::Some | PatternCtorTag::None => MatchFamily::Option,
+            PatternCtorTag::False | PatternCtorTag::True => MatchFamily::Bool,
+        }
+    }
+
+    fn is_payload(self) -> bool {
+        matches!(
+            self,
+            PatternCtorTag::Left | PatternCtorTag::Right | PatternCtorTag::Some
+        )
+    }
+
+    fn as_payload_tag(self) -> &'static str {
+        match self {
+            PatternCtorTag::Left => "left",
+            PatternCtorTag::Right => "right",
+            PatternCtorTag::Some => "some",
+            PatternCtorTag::None | PatternCtorTag::False | PatternCtorTag::True => {
+                unreachable!("Nullary constructors do not carry payloads")
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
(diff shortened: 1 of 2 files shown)


metalurgical:
Resolves #250 
Closes #250

KyrylR:
Quite tough to review, as a single commit contains too many changes

The main reason for having smaller commits is to isolate possible regressions, so when the commit history follows the rule of a "logical sequence of clearly defined independent changes" it is easier to debug

For example, if at some point a bug/regression is discovered, I can write a test and find the exact commit that introduced it
But if this regression happened in a commit like this, it would be much harder to say where exactly

This is why I have to ask you to split this MR into smaller commits

KyrylR:
> Quite tough to review, as a single commit contains too many changes
> 
> The main reason for having smaller commits is to isolate possible regressions, so when the commit history follows the rule of a "logical sequence of clearly defined independent changes" it is easier to debug
> 
> For example, if at some point a bug/regression is discovered, I can write a test and find the exact commit that introduced it But if this regression happened in a commit like this, it would be much harder to say where exactly
> 
> This is why I have to ask you to split this MR into smaller commits

I expect to see a way smaller commits

metalurgical:
> > Quite tough to review, as a single commit contains too many changes
> > The main reason for having smaller commits is to isolate possible regressions, so when the commit history follows the rule of a "logical sequence of clearly defined independent changes" it is easier to debug
> > For example, if at some point a bug/regression is discovered, I can write a test and find the exact commit that introduced it But if this regression happened in a commit like this, it would be much harder to say where exactly
> > This is why I have to ask you to split this MR into smaller commits
> 
> I expect to see a way smaller commits

I understand the concern around splitting the PR, and in many cases I would agree that smaller, separated commits improve reviewability.

In this case, however, the core of the change is concentrated in:
- https://github.com/BlockstreamResearch/SimplicityHL/pull/275/changes#diff-5583e398a8deec11b274d9965eff8b5ade5226c7a020ca214a27d7d07dcc8a29L1704-L1728
- https://github.com/BlockstreamResearch/SimplicityHL/pull/275/changes#diff-5583e398a8deec11b274d9965eff8b5ade5226c7a020ca214a27d7d07dcc8a29L1734-R2611

Everything else follows from this and consists of new logic rather than modifications to existing behavior.

It has been structured as a single unit so it can be reviewed in a working state.

Splitting this into separate commits would likely introduce more overhead by requiring context to be reconstructed across commits and would be simpler to review as it currently is.

Additionally, I did add the example case that was presented on the original issue as a test case as well for completeness (in a separate commit).

Happy to discuss it further if that would help.

apoelstra:
The first mega-commit has stuff like typo fixes and changes from `std::fmt::Debug` to `fmt::Debug`. At the very least these can be split out.

The tests module can be moved to a separate commit, in front of the others, with the tests that don't yet pass marked `should_fail` or `ignore`. This would make the tests more effective as regressions, and also make it clearer to the reader what changes are happening.

It's possible, though maybe not very helpful, to move the new types into their own commits and then the changes that *use* them will be isolated. As you say, this may just remove context. But OTOH, if you use multiple commits to introduce the new types, then it's easier to see logically what builds on what.

I appreciate the hint as to where the changes "start".

metalurgical:
> The first mega-commit has stuff like typo fixes and changes from `std::fmt::Debug` to `fmt::Debug`. At the very least these can be split out.
> 
> The tests module can be moved to a separate commit, in front of the others, with the tests that don't yet pass marked `should_fail` or `ignore`. This would make the tests more effective as regressions, and also make it clearer to the reader what changes are happening.
> 
> It's possible, though maybe not very helpful, to move the new types into their own commits and then the changes that *use* them will be isolated. As you say, this may just remove context. But OTOH, if you use multiple commits to introduce the new types, then it's easier to see logically what builds on what.
> 
> I appreciate the hint as to where the changes "start".

Understood. Will break it up.

metalurgical:
Have realized that this can be greatly simplified, converting to draft for now.

KyrylR:
Closed in favor of https://github.com/BlockstreamResearch/SimplicityHL/pull/376

Issue #252, opened 2026-03-23
SimplicityHL doesn't understand T::MIN and T::MAX


schoen:
### Project version

0.4.1

### Project

core

### What happened?

The jet documentation (as extracted from SimplicityHL source code) refers to `u1::MIN`, `u2::MIN`, ... `u64::MIN`, 
as well as `u1::MAX`, `u2::MAX`, ... `u64::MAX`. One might also expect `u128::MAX` and `u256::MAX` to exist.

However, the SimplicityHL compiler doesn't currently recognize any of these notations and we can't successfully use them in a program. They might arguably refer to compiler internals, but, if so, they probably shouldn't be mentioned in user-facing documentation.

### Minimal reproduction steps

```
fn main(){
   let x: u8 = u8::MIN;
}
```

KyrylR:
@stringhandler Does it make sense to open PR with your solution? (so we can compare both approaches)

stringhandler:
I think it's better to wait for a standard lib and then see if it's still necessary after

PR #277, opened 2026-04-07
feat: add support for unsigned integer MIN and MAX type bounds

metalurgical:
Introduce parser support for unsigned integer constants (MIN, MAX) across all variants.

Can be further extended easily to include other type specific constants like ZERO, ONE, etc in future.

Extensible for future types that need specific constants as well, without interfering with existing constant types.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -249,6 +249,28 @@ impl TypeAlias {
 
 impl_eq_hash!(TypeAlias; name, ty);
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+pub enum UIntBound {
+    Min,
+    Max,
+}
+
+impl UIntBound {
+    pub const fn as_str(self) -> &'static str {
+        match self {
+            Self::Min => "MIN",
+            Self::Max => "MAX",
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+pub enum TypeBound {
+    UInt(UIntType, UIntBound),
+}
+
 /// An expression is something that returns a value.
 #[derive(Clone, Debug)]
 pub struct Expression {
@@ -339,6 +361,8 @@ pub enum SingleExpressionInner {
     Binary(Binary),
     /// Hexadecimal string literal.
     Hexadecimal(Hexadecimal),
+    /// Constants of a type (e.g. MAX, MIN)
+    TypeBound(TypeBound),
     /// Witness value.
     Witness(WitnessName),
     /// Parameter value.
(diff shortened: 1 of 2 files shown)


metalurgical:
Resolves #252 
Closes #252

KyrylR:
Thanks for the contribution!

Overall it LGTM, but there is a nuance with errors

src/parse.rs
```diff
@@ -1606,6 +1632,38 @@ impl SingleExpression {
 
         let match_expr = Match::parser(expr.clone()).map(SingleExpressionInner::Match);
 
+        let type_bound = Identifier::parser()
+            .then_ignore(just(Token::DoubleColon))
+            .then(Identifier::parser())
+            .validate(|(lhs, rhs), e, emit| {
```
KyrylR: This thing means that any ident::ident shape is consumed first, like: 
* u8::MAX
* foo::bar
* witnes::PK

The problem is the .validate(...) step does not reject the parse when the identifiers are wrong. It emits errors, but still returns a TypeBound node using fallback dummy values. So e.g witnes::PK never gets a chance to “fall through” to the normal parser path

KyrylR:
So now with this example:
```
fn main() { let pk: Pubkey = witnes::PK; }
```

Error changed from:
> Expected ';', found '::'

to things like:

> Expected unsigned integer type before `::`
> Expected `MIN` or `MAX` after unsigned integer type

KyrylR:
image

metalurgical:
> So now with this example:
> ```
> fn main() { let pk: Pubkey = witnes::PK; }
> ```
> 
> Error changed from:
> > Expected ';', found '::'
> 
> to things like:
> 
> > Expected unsigned integer type before `::`
> > Expected `MIN` or `MAX` after unsigned integer type

Thanks for pointing this out. +1

KyrylR:
Nevertheless, would like to hear your opinion on this

metalurgical:
> Nevertheless, would like to hear your opinion on this

I had incorrectly assumed this case would have been caught by existing tests. Additional tests have been added around this and corrected the implementation here.

KyrylR (approved):

KyrylR:
just for context, waiting for the second review before merging

stringhandler (commented):
Long time, Strider, I hope you are well. I think it would actually be better to model this properly. It's a Const or a Call at best, rather than a new unique language feature. I would rather say we should treat it as an Ident on the lexer, a Constant(ConstantName) on the parser and AST level, kind of how we do Call. It could maybe even be written as a call.

metalurgical:
> Long time, Strider, I hope you are well. 

@stringhandler It has been and likewise.

> I think it would actually be better to model this properly. It's a Const or a Call at best, rather than a new unique language feature. I would rather say we should treat it as an Ident on the lexer, a Constant(ConstantName) on the parser and AST level, kind of how we do Call. It could maybe even be written as a call.

The current approach was intentionally limited to address the immediate issue with minimal changes, while leaving room for future extension if needed. 

I agree that modelling this as part of a more general language construct would be both sound and a cleaner long-term solution.

The current approach is contained such that it does not prevent transitioning to that model in future.

KyrylR (changes requested):
Just to clarify the confusion, we are taking time to asses other possibilities

KyrylR:
> Just to clarify the confusion, we are taking time to asses other possibilities

The main question for us to tackle is how far away we are going with the support of constants/methods for other types as well

KyrylR:
Commented here: https://github.com/BlockstreamResearch/SimplicityHL/issues/252#issuecomment-4405238869

stringhandler:
@metalurgical can we close this?

apoelstra:
bbf4cbc5c4f39b87ed2dd83efd32427417fd928a needs rebase

metalurgical:
> @metalurgical can we close this?

Unknown, a previous comment indicated other possibilities were to be explored first.

KyrylR:
Closing this for the time being 

We can reopen it after the further discussion in the #252

PR #279, opened 2026-04-08
Implement Linearization for dependency resolution

LesterEvSe:
This PR introduces the Linearization algorithm to manage module dependencies. It handles complex dependency topologies, detects cycles, and ensures strict, deterministic load ordering.

### DIFF
--- a/src/driver/linearization.rs
+++ b/src/driver/linearization.rs
@@ -1,0 +1,184 @@
+//! Computes the bottom-up load order of dependencies using a DFS.
+//!
+//! # Architectural Note: Why pure DFS instead of C3?
+//! Unlike OOP languages that require C3 Linearization to resolve complex method
+//! overriding (MRO), SimplicityHL module imports rely on namespacing and aliasing.
+//! Because we do not need to enforce strict local precedence, a standard post-order
+//! DFS is a better option.
+
+// TODO: Remove this once the code is actively used.
+#![allow(dead_code)]
+use std::collections::HashSet;
+use std::fmt;
+
+use crate::driver::DependencyGraph;
+
+/// This is a core component of the [`DependencyGraph`].
+impl DependencyGraph {
+    /// Returns the deterministic, BOTTOM-UP load order of dependencies.
+    pub fn linearize(&self) -> Result<Vec<usize>, LinearizationError> {
+        let mut visited = HashSet::new();
+        let mut visiting = Vec::new();
+        let mut order = Vec::new();
+
+        self.dfs_linearize(0, &mut visited, &mut visiting, &mut order)?;
+
+        Ok(order)
+    }
+
+    /// Core recursive Post-Order DFS for topological sorting.
+    ///
+    /// - **Visited Set (`visited`):** Prevents processing shared dependencies multiple times (solves diamonds).
+    /// - **Cycle Detection (`visiting`):** Tracks the current path stack to catch infinite loops.
+    /// - **Order List (`order`):** Accumulates the deterministic load order bottom-up.
+    fn dfs_linearize(
+        &self,
+        module: usize,
+        visited: &mut HashSet<usize>,
+        visiting: &mut Vec<usize>,
+        order: &mut Vec<usize>,
+    ) -> Result<(), LinearizationError> {
+        // If we have already fully processed this module, skip it (Diamond Deduplication)
+        if visited.contains(&module) {
+            return Ok(());
+        }
+
+        if let Some(cycle_start) = visiting.iter().position(|&m| m == module) {
+            return Err(LinearizationError::CycleDetected(
+                visiting[cycle_start..]
+                    .iter()
+                    .map(|&id| self.modules[id].source.str_name())
+                    .collect(),
+            ));
+        }
+
+        visiting.push(module);
+
+        let parents = self
+            .dependencies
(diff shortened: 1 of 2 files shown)


KyrylR:
Is this up to date: https://github.com/BlockstreamResearch/SimplicityHL/pull/238?

LesterEvSe:
> Is this up to date: #238?

Yes, but it is much more optimized and has been refactored extensively

src/driver/c3_linearization.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        for &parent in parents {
+            seqs.push(self.linearize_rec(parent, memo, visiting)?);
+        }
+
+        let mut result = vec![module];
+        let merged = c3_merge(seqs).map_err(|conflicts| C3Error::InconsistentLinearization {
+            module: self.modules[module].source.str_name(),
+            conflicts: self.format_conflict_names(conflicts),
+        })?;
```
KyrylR: From what I understand the C3 should merge each parent linearization plus the raw parents list, otherwise the direct parent order is never part of the constraints. With the current code, parents [A, B] and linearizations [A] / [B, A] produce [B, A], even though that order contradicts the explicit parent list.

src/driver/c3_linearization.rs
```diff
@@ -0,0 +1,222 @@
...
+    #[test]
+    fn test_c3_detects_cycle() {
+        let (graph, _, _dir) = setup_graph(vec![
+            ("main.simf", "use lib::A::entry;"),
+            ("libs/lib/A.simf", "use lib::B::func;"),
+            ("libs/lib/B.simf", "use lib::A::func;"),
+        ]);
+
+        let order = graph.c3_linearize();
+        matches!(order, Err(C3Error::CycleDetected(_)));
```
KyrylR: `matches!` just returns a boolean... the result is discarded here

src/driver/c3_linearization.rs
KyrylR:
```
    #[test]
    fn test_c3_rejects_conflicting_direct_parent_order() {
        // main declares parents in the order [A, B].
        // But B itself depends on A, so B's linearization is [B, A].
        //
        // True C3 must merge:
        //   [A]
        //   [B, A]
        //   [A, B]   <- the raw parent list
        //
        // That merge has no valid head, so this must be rejected as inconsistent.
        let (graph, _, _dir) = setup_graph(vec![
            ("main.simf", "use lib::A::foo; use lib::B::bar;"),
            ("libs/lib/A.simf", ""),
            ("libs/lib/B.simf", "use lib::A::foo;"),
        ]);

        let order = graph.c3_linearize();

        assert!(
            matches!(order, Err(C3Error::InconsistentLinearization { .. })),
            "expected inconsistent linearization, got {order:?}"
        );
    }
```

KyrylR (changes requested):

src/driver/c3_linearization.rs
LesterEvSe: Fixed and added a new test. I accidentally missed adding the parents' order to the sequences when I refactored the code

src/driver/c3_linearization.rs
```diff
@@ -0,0 +1,251 @@
+// TODO: Remove this once the code is actively used.
+#![allow(dead_code)]
+use std::{collections::HashMap, fmt};
+
+use crate::driver::DependencyGraph;
+
+impl DependencyGraph {
```
KyrylR: Give a link in the rust doc to the parent structure pls

KyrylR:
```
    #[test]
    fn test_c3_allows_valid_parent_chain() {
        // main imports A, then B; B itself imports A.
        let (graph, ids, _dir) = setup_graph(vec![
            ("main.simf", "use lib::A::foo; use lib::B::bar;"),
            ("libs/lib/A.simf", ""),
            ("libs/lib/B.simf", "use lib::A::foo;"),
        ]);

        let order = graph
            .c3_linearize()
            .expect("valid dependency DAG should linearize successfully");

        let main_id = ids["main"];
        let a_id = ids["A"];
        let b_id = ids["B"];

        assert_eq!(order, vec![main_id, b_id, a_id]);
    }
```

Now the valid import sequence is rejected

KyrylR:
Also, is it expected that the order is top-down?

LesterEvSe:
> Also, is it expected that the order is top-down?

I have decided to move `order.reverse()` directly to the `c3_linearize()` function. So now it is bottom-up at this stage

src/driver/c3_linearization.rs
LesterEvSe: Update: I was mistaken in my previous comment. The `test_c3_allows_valid_parent_chain()` proves that we cannot enforce strict C3 parent order without breaking valid module imports. I have retained the 'Dependency-Driven' C3 approach and added a top-level docstring to formally document this architectural decision

LesterEvSe: Update 2: Taking this logic a step further—since we explicitly do not want to enforce strict local precedence for module imports, carrying the C3 is actually unnecessary. A standard post-order DFS provides this as well, but without overhead

KyrylR (approved):
ACK 9ca524ce6aeeeae7b03d3808654efae57415b33d

PR #280, opened 2026-04-10
Implement Program AST generation in the Driver

LesterEvSe:
This PR integrates the previously written driver functionality to construct the unified `Program` AST structure.
Also, some errors were corrected here. For now, it's difficult to create effective error handling with the current type of handler, so this will be a task outside of these series of module PRs.

### DIFF
--- a/src/driver/resolve_order.rs
+++ b/src/driver/resolve_order.rs
@@ -1,0 +1,291 @@
+use std::collections::{BTreeSet, HashMap};
+use std::sync::Arc;
+
+use crate::driver::DependencyGraph;
+use crate::error::{Error, ErrorCollector, RichError, Span};
+use crate::impl_eq_hash;
+use crate::parse::{self, Visibility};
+use crate::resolution::CanonPath;
+
+/// The final, flattened representation of a SimplicityHL program.
+///
+/// This struct holds the fully resolved sequence of items, paths, and scope
+/// resolutions, ready to be passed to the next stage of the compiler.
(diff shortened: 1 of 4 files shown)


src/driver/resolve_order.rs
```diff
@@ -0,0 +1,313 @@
...
+                if matches!(elem, parse::Item::Module) {
+                    continue;
+                }
+
+                // 3. Handle Types & Functions
+                let (name, vis) = match elem {
+                    parse::Item::TypeAlias(a) => (a.name().as_inner(), a.visibility()),
+                    parse::Item::Function(f) => (f.name().as_inner(), f.visibility()),
+                    _ => unreachable!(),
+                };
```
KyrylR: Combine those two instead of using unreachable

src/driver/resolve_order.rs
```diff
@@ -0,0 +1,313 @@
...
+
+        if handler.has_errors() {
+            None
+        } else {
+            Some(Program {
+                items: items.into(),
+                files: construct_resolved_file_array(&self.paths, &resolutions),
+                span: *self.modules[0].parsed_program.as_ref(),
+            })
+        }
```
KyrylR: Let's not use if-else

KyrylR (changes requested):

KyrylR (approved):
ACK 4b0b8347fb301fdb1881f699ecc0e77a6602a7d5; code review

PR #282, opened 2026-04-13
feat: add out of order function calls

stringhandler:
Adds a double parsing step for analysis in order to allow functions to appear later in the source code.

# Motivation:
With the upcoming library references PRs I created a test standard library [here](https://github.com/stringhandler/simplicity-std). It quickly became clear that it would be difficult to maintain the ordering of function calls so that they were always declared before they were called. This PR removes that restriction by doing a few prechecks before analyzing the parsed tree.

Allowing functions to be declared after they are called now means that extra checks must be done to prevent recursion. Since `main` is treated a little differently, I have not allowed a function `main` to be called.

# Update
As noted by @KyrylR, types are also allowed to be referenced out of order.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -1,5 +1,5 @@
 use std::collections::hash_map::Entry;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::num::NonZeroUsize;
 use std::str::FromStr;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


KyrylR:
It is not mentioned in the description, but with this PR types could be also declared out of order, reproduction test:
```
    #[test]
    fn type_alias_must_be_declared_before_use() {
        compile_err(
            r#"fn main() {
    let x: Later = 32;
    assert!(jet::eq_32(x, 32));
}

type Later = u32;"#,
            "Type alias `Later` is not defined",
        );
    }
```

Also this PR introduces regression over duplicated functions, here is the test:
```
    #[test]
    fn duplicate_helper_function_redefinition_rejected() {
        compile_err(
            r#"fn foo() -> bool {
    true
}

fn foo() -> bool {
    false
}

fn main() {
    assert!(foo());
}"#,
            "Function `foo` was defined multiple times",
        );
    }
```

KyrylR (changes requested):
See comment above

stringhandler:
Thanks, will fix

KyrylR:
needs rebase

stringhandler:
Will probably close and redo a bit later. The code is so fluid right now.

PR #284, opened 2026-04-15
Change AST to match the functionality of the Driver

LesterEvSe:
- Adapt the AST to enable driver functionality.
- Add architecture notes about using "main" as an alias.
- Add constants for the "main" function and the first module in the `DependencyGraph` to prevent the usage of magic numbers.

AST currently has some issues with processing source files properly, so they are not used and have been deleted along with the `ResolvedFile` structure.
I hope this will be fixed when the `ErrorCollector` is recoded to support multiple files better.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -9,6 +9,7 @@ use miniscript::iter::{Tree, TreeLike};
 use simplicity::jet::Elements;
 
 use crate::debug::{CallTracker, DebugSymbols, TrackedCallName};
+use crate::driver::{FileScoped, SymbolTable, MAIN_MODULE, MAIN_STR};
 use crate::error::{Error, RichError, Span, WithSpan};
 use crate::num::{NonZeroPow2Usize, Pow2Usize};
 use crate::parse::MatchPattern;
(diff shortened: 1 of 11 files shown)


src/driver/mod.rs
```diff
@@ -28,7 +28,7 @@
 //! the dependency graph construction.
 
 mod linearization;
-mod resolve_order;
+pub mod resolve_order;
```
```suggestion
pub(crate) mod resolve_order;
```
?

KyrylR:
```cargo rustc --manifest-path fuzz/Cargo.toml --bin compile_parse_tree -- --cfg fuzzing``` is failing

**Comment by KyrylR** [comment]
Could we document that `main` cannot be used as an alias?

File: src/ast.rs [hunk 2]
```diff
@@ -551,6 +586,10 @@ impl Scope {
     pub fn push_main_scope(&mut self) {
         assert!(!self.is_main, "Already inside main function");
         assert!(self.is_topmost(), "Current scope is not topmost");
+        assert_eq!(
+            self.file_id, MAIN_MODULE,
+            "The 'main' function must be defined in the entry point file"
+        );
```
KyrylR: Not sure if it is a good idea to panic here

KyrylR (approved):
ACK 0fc6ca3f58da2fbc881edd330cf010a8fcbae5e3

PR #287, opened 2026-04-17
Fix: resolution structs and import aliases to support separated namespaces

LesterEvSe:
Currently, SimplicityHL allows separated namespaces. A user should be able to use the exact same name for both a type and a function in the same scope. For example:

``` Rust
type temp = u32;
fn temp() {}
```

The previous alias tracking functionality broke this behavior during the use import resolution phase.
So, This PR refactors the alias architecture and the `process_use_item` logic.

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -136,6 +136,7 @@ pub struct DependencyGraph {
     /// This serves as the exact inverse of the `lookup` map.
     paths: Vec<CanonPath>,
 
+    // TODO: Consider to optimising this with `Vec` instead of `HashMap`
     /// The Adjacency List: Defines the Directed acyclic Graph (DAG) of imports.
     ///
     /// The Key (`usize`) is the ID of a "Parent" module (the file doing the importing).
--- a/doc/architecture.md
+++ b/doc/architecture.md
@@ -5,3 +5,17 @@ paths during import parsing. However, in our architecture, the prefix before
 the first `::` in a `use` statement is always an dependency root path. Since all
 dependency root paths are unique and strictly bound to specific paths, the resolver
 can always unambiguously resolve the path without needing relative pointers.
+
+# Architecture Note: Namespace Separation in SimplicityHL
+
+SimplicityHL uses distinct namespaces for types and values. This allows the same identifier to refer to different things depending on where it appears in the code — the compiler determines the correct interpretation from syntactic context, with no risk of collision.
+
+``` Rust
+fn foo() -> bool { false }
+type foo = bool;
+
+fn main() {
+    let x: foo = false; // type namespace  - type alias
+    assert!(foo());     // value namespace - function
+}
+```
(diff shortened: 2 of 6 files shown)


KyrylR:
Was this thing allowed in simplicity 0.4.1?

KyrylR:
> Was this thing allowed in simplicity 0.4.1?

I was able to compile this:
```rust
fn foo() -> bool { false }
type foo = bool;

fn main() {
    let x: foo = false;
    assert!(foo());
}
```

src/driver/mod.rs
```diff
@@ -136,6 +136,7 @@ pub struct DependencyGraph {
     /// This serves as the exact inverse of the `lookup` map.
     paths: Vec<CanonPath>,
 
+    // NOTE: Consider to optimising this with `Vec` instead of `HashMap`
```
```suggestion
    // TODO: Consider to optimising this with `Vec` instead of `HashMap`
```

KyrylR:
At the same time the following test passes:
```
    #[test]
    fn test_private_alias_error_does_not_mask_duplicate_function_import() {
        let (graph, _ids, _dir) = setup_graph(vec![
            ("libs/lib/A.simf", "pub fn foo() {}"),
            ("libs/lib/B.simf", "pub fn foo() {} type foo = u32;"),
            ("main.simf", "use lib::A::foo; use lib::B::foo;"),
        ]);

        let mut error_handler = ErrorCollector::new();
        let program_option = graph.linearize_and_build(&mut error_handler).unwrap();
        let errors = error_handler.to_string();

        assert!(
            program_option.is_none(),
            "build should fail when a second import reuses the function name `foo`"
        );
    }
  ```

KyrylR:
ACK 763086bf724c64aa880e06678e7ba183317ed961

So I would mention the behavior above in the md doc

LesterEvSe:
> > Was this thing allowed in simplicity 0.4.1?
> 
> I was able to compile this:
> 
> ```rust
> fn foo() -> bool { false }
> type foo = bool;
> 
> fn main() {
>     let x: foo = false;
>     assert!(foo());
> }
> ```

Yes, SimplicityHL 0.4.1 allows this

LesterEvSe:
> At the same time the following test passes:
> 
> ```
>     #[test]
>     fn test_private_alias_error_does_not_mask_duplicate_function_import() {
>         let (graph, _ids, _dir) = setup_graph(vec![
>             ("libs/lib/A.simf", "pub fn foo() {}"),
>             ("libs/lib/B.simf", "pub fn foo() {} type foo = u32;"),
>             ("main.simf", "use lib::A::foo; use lib::B::foo;"),
>         ]);
> 
>         let mut error_handler = ErrorCollector::new();
>         let program_option = graph.linearize_and_build(&mut error_handler).unwrap();
>         let errors = error_handler.to_string();
> 
>         assert!(
>             program_option.is_none(),
>             "build should fail when a second import reuses the function name `foo`"
>         );
>     }
> ```

It is okay too. I added this test and explain why it passed

KyrylR (commented):
ACK https://github.com/BlockstreamResearch/SimplicityHL/pull/287/commits/feb901e58b675805711953838fc528c6fe0d04aa

Issue #291, opened 2026-04-22
Bug: Execution reached a pruned branch for certain fold combinations


stringhandler:
### Project version

0.5.0

### Project

compiler

### What happened?

I think this has something to do with a node being shallow cloned instead of deep cloning. I used claude to generate some tests for various combinations of list lengths. The lengths 0,2,3 work for `List<T, 4>`, but length 1 fails.

### Minimal reproduction steps

Create unit test with the following:

```rust 
#[cfg(test)]
mod tests {
    use std::borrow::Cow;

    use crate::{tests::TestCase, WitnessValues};

    // Helper: fold with addition, f(elt, acc) = acc + elt
    const ADD_FN: &str = r#"fn add(elt: u32, acc: u32) -> u32 {
    let (_, sum): (bool, u32) = jet::add_32(elt, acc);
    sum
}
"#;

    // Helper: fold that ignores the accumulator and returns the element.
    // When folded left-to-right over a non-empty list, the final result equals the LAST element.
    const LAST_FN: &str = r#"fn last(elt: u32, _acc: u32) -> u32 {
    elt
}
"#;


    fn make_prog(fns: &str, body: &str) -> String {
        format!("{fns}\nfn main() {{\n{body}\n}}\n")
    }

    fn run(prog: &str) {
        TestCase::program_text(Cow::Owned(prog.to_owned()))
            .with_witness_values(WitnessValues::default())
            .assert_run_success();
    }


    #[test]
    fn list_fold_one_element_bound4() {
        // Partition: head-block empty, tail = [5].
        run(&make_prog(
            ADD_FN,
            r#"    let list: List<u32, 4> = list![5];
    let result: u32 = fold::<add, 4>(list, 0);
    assert!(jet::eq_32(result, 5));"#,
        ));
    }

  #[test]
    fn list_fold_five_elements_bound8() {
        // Partition: outer-head=[1,2,3,4], outer-tail partition [5].
        run(&make_prog(
            ADD_FN,
            r#"    let list: List<u32, 8> = list![1, 2, 3, 4, 5];
    let result: u32 = fold::<add, 8>(list, 0);
    assert!(jet::eq_32(result, 15));"#,
        ));
    }

```

## Expected
Tests pass

## Actual
```

---- compile::tests::list_fold_one_element_bound4 stdout ----

thread 'compile::tests::list_fold_one_element_bound4' (17336) panicked at src\lib.rs:462:31:
[...]

Issue #288, opened 2026-04-21
Why does List<T, N> hold a maximum of N-1 elements?


ikripaka:
While implementing macros for Simplex, I noticed that the List<T, N> type can hold strictly fewer elements than its specified size bound. 
For example, if we declare a list with a size limit of 16, it can only hold a maximum of 15 elements.

https://docs.simplicity-lang.org/simplicityhl-reference/type/
https://github.com/BlockstreamResearch/SimplicityHL/blob/aeeb37ee057a242cdb5b5ef166d5a15539b9f1f5/src/value.rs#L955
https://github.com/BlockstreamResearch/SimplicityHL/blob/aeeb37ee057a242cdb5b5ef166d5a15539b9f1f5/src/array.rs#L179

I'm curious about the technical reasoning behind this design choice. Are you inserting the head of the list as the 0th element, resulting in the truncated capacity?
Or a binary tree uses a recursive cutoff, by using 2^i elements in each leaf? 
Or is this purely related to how lists are encoded for contract stack execution?

From a developer's perspective, it feels counterintuitive that a list bound of N forces an "off-by-one" capacity limit.

stringhandler:
I also noticed this to be a bit odd. There is an explanation on [the types documentation](https://docs.simplicity-lang.org/simplicityhl-reference/type/), under Lists:

> Lists hold a variable number of elements of the same type. This is similar to [Rust vectors](https://doc.rust-lang.org/std/vec/struct.Vec.html), but SimplicityHL doesn't have a heap. In SimplicityHL, lists exist on the stack, which is why the maximum list length is bounded.

> <2-lists hold fewer than 2 elements, so zero or one element. <4-lists hold fewer than 4 elements, so zero to three elements. <8-lists hold fewer than 8 elements, so zero to seven elements. And so on. For technical reasons, the list bound is always a power of two. The bound 1 is not supported, because it would only allow empty lists, which is useless.


> Lists are different from arrays: List values hold a variable number of elements. Array values hold a fixed number of elements.

> On the blockchain, you pay for every byte that you use. If you use an array, then you pay for every single element. For example, values of type [u8; 512] cost roughly as much as 512 many u8 values. However, if you use a list, then you only pay for the elements that you actually use. For example, the type List<u8, 512> allows for up to 511 elements. If you only use three elements list![1, 2, 3], then you pay for exactly three elements. You don't pay for the remaining 508 unused elements

This does explain it, but I also find the notation counter-intuitive

[...]

PR #292, opened 2026-04-22
fix: fix list_fold pruning too soon

stringhandler:
NOTE: This code was generated by Claude. 

I asked Claude to add some covering tests for `list_fold` using different lengths. It helpfully identified a bug and fixed it. To be honest, I do not understand this portion of the code well enough to review it properly, but I'm posting it here as a draft PR in case we can review it and understand the problem.

The tests it generated certainly fail with the old code, as shown in #291 

Related to #288

### DIFF
--- a/src/compile/mod.rs
+++ b/src/compile/mod.rs
@@ -486,26 +486,37 @@ fn list_fold<'brand>(
         f_array: &ProgNode<'brand>,
         f_fold: &ProgNode<'brand>,
     ) -> Result<ProgNode<'brand>, simplicity::types::Error> {
-        /* (fold f)_(n + 1) :  E<2^(n + 1) × A → A
+        /* (fold f)_(n + 1) :  E^<2^(n + 1) × A → A
          * (fold f)_(n + 1) := OOH ▵ (OIH ▵ IH);
-         *                     case (drop (fold f)_n)
-         *                          ((IOH ▵ (OH ▵ IIH; f_n)); (fold f)_n)
+         *                     (IOH ▵ case IIH (OH ▵ IIH; f_n));
+         *                     (fold f)_n
+         *
+         * Uses f_fold exactly once (no DAG sharing) to avoid incorrect pruning.
          */
         let ctx = f_array.inference_context();
+        // Rearrange input: ((1+E^n) × E^<n) × A → (1+E^n, (E^<n, A))
         let case_input = ProgNode::o()
             .o()
             .h(ctx)
             .pair(ProgNode::o().i().h(ctx).pair(ProgNode::i().h(ctx)));
-        let case_left = ProgNode::drop_(f_fold);
 
-        let f_n_input = ProgNode::o().h(ctx).pair(ProgNode::i().i().h(ctx));
-        let f_n_output = f_n_input.comp(f_array)?;
-        let fold_n_input = ProgNode::i().o().h(ctx).pair(f_n_output);
-        let case_right = fold_n_input.comp(f_fold)?;
+        // acc' = case IIH (OH ▵ IIH; f_n)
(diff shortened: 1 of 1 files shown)


apoelstra:
I'm extremely skeptical of this fix. ~~If you remove debug symbols and pruning in the test code, you get errors like~~

```
thread 'compile::tests::list_fold_one_element_bound4' (1233321) panicked at /home/apoelstra/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/simplicity-lang-0.7.0/src/bit_machine/mod.rs:80:9:
```

~~which seems just crazy. I agree with your intuition that we're somehow producing a corrupted program tree, maybe by cloning things we shouldn't. This PR seems to change the fold logic, which is unlikely to improve anything except by accident.~~

Never mind, I messed up my bit-machine construction. I am still skeptical of this fix.

apoelstra:
You can disable pruning in the test code with

```diff
diff --git a/src/lib.rs b/src/lib.rs
index 6e9bca7..832e1f6 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -451,9 +451,12 @@ pub(crate) mod tests {
         fn run(self) -> Result<(), simplicity::bit_machine::ExecutionError> {
             let env = dummy_env::dummy_with(self.lock_time, self.sequence, self.include_fee_output);
             let pruned = self.program.redeem().prune(&env)?;
-            let mut mac = BitMachine::for_program(&pruned)
+            println!("redeem  {:?}", self.program.redeem().display_expr());
+            println!("pruned  {:?}", pruned.display_expr());
+
+            let mut mac = BitMachine::for_program(&self.program.redeem())
                 .expect("program should be within reasonable bounds");
-            mac.exec(&pruned, &env).map(|_| ())
+            mac.exec(&self.program.redeem(), &env).map(|_| ())
         }

         pub fn assert_run_success(self) {
```

Then the `bound4` test passes. So I suspect this is actually a bug in rust-simplicity's pruning logic. I *think* it should be true that for any `Redeem` program you can find:

1. If `BitMachine::exec` succeeds on that program...
2. then `Redeem::prune` should succeed on that program...
3. and `BitMachine::exec` should succeed on the output of pruning

assuming the tx environment is the same in all cases (which it is here, though this doesn't even matter since we're not using any tx primitives).

I would suggest adding a fuzztest to rust-simplicity trying to find simple cases where the above fails.

stringhandler:
I agree. There may be a bug here, but more importantly, the pruning should fail first.

I'll add the fuzz test

stringhandler:
Just an update here. I did look into creating the fuzzing test today. The main difficulty is that very few programs generated in the fuzzers are valid enough to even pass the BitMachine a first time (around 5-6%). I switched strategies to try generate better programs in SimplicityHL, but I saw that it also doesn't generate that many valid programs. I created pr #302 for now but it doesn't often generate programs that are interesting enough to even be pruned. 

I plan to continue this next week though and see if I can improve the programs generated in fuzzing to the point that this test is getting actual hits.

apoelstra:
@stringhandler I think you should be able to create the fuzztest in `rust-simplicity` directly using Simplicity programs, and bypass SimplicityHL entirely. My feeling is that the SimplicityHL `fold` stuff is a red herring; I think we have a general pruning bug and it just happens that `fold` is where the HL compiler hits it.

Though your work in #302 is independently valuable.

apoelstra:
I am traveling for the next week but will be back at work after that and can probably help out.

stringhandler:
Yeah agree with it being a pruning bug. I'll continue to try get more realistic programs in rust-simplicity and SimplicityHL

stringhandler:
I believe this is the actual fix https://github.com/BlockstreamResearch/rust-simplicity/pull/362

PR #303, opened 2026-04-27
Crate keyword

### LINKED ISSUE #294: Introduce `crate` keyword for local imports
### Project

driver

### Describe the feature

Currently, the compiler processes all dependencies through the `--dep` flag without distinguishing between local files and external libraries. So it would be better to introduce `crate` keyword to explicitly mark local program files, separating them from external dependencies.

### Motivation
- **Prevents a breaking change in future versions.** Adding this distinction later would require rewriting all existing import paths.
- **Improves error reporting.** It would be easier for the error handler to tell whether an unresolved path is a missing local file or a missing external dependency.
- **Improves code readability.** For example, `crate::utils::foo` immediately signals local code, while `mathlib::utils::foo` signals an external library.

Sdoba16:
Added support of crate keyword for local dependencies.
Added check that local files are dependent only with crate keyword.
Small fix for test_utils.
Closes #294

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1461,9 +1461,9 @@ impl ChumskyParse for UseDecl {
             .or_not()
             .map(Option::unwrap_or_default);
 
-        // Parse the base path prefix (e.g., `dependency_root_path::file::` or `dependency_root_path::dir::file::`).
+        // Parse the base path prefix (e.g., `dependency_root_path::file::`, `dependency_root_path::dir::file::`, or `crate::dir::file::`).
         // We require at least 2 segments here because a valid import needs a minimum
-        // of 3 items total: the dependency_root_path, the file, and the specific item/function.
+        // of 3 items total: the dependency root path (or `crate`), the file, and the specific item/function.
         let first_segment = select! {
             Token::Ident(ident) => Identifier::from_str_unchecked(ident),
             Token::Crate => Identifier::from_str_unchecked(CRATE_STR),
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -17,7 +17,7 @@ use chumsky::{extra, select, IterParser, Parser};
 use either::Either;
 use miniscript::iter::{Tree, TreeLike};
 
-use crate::driver::MAIN_MODULE;
+use crate::driver::{CRATE_STR, MAIN_MODULE};
 use crate::error::ErrorCollector;
 use crate::error::{Error, RichError, Span};
 use crate::impl_eq_hash;
@@ -1464,11 +1464,25 @@ impl ChumskyParse for UseDecl {
         // Parse the base path prefix (e.g., `dependency_root_path::file::` or `dependency_root_path::dir::file::`).
         // We require at least 2 segments here because a valid import needs a minimum
         // of 3 items total: the dependency_root_path, the file, and the specific item/function.
-        let path = Identifier::parser()
+        let first_segment = select! {
+            Token::Ident(ident) => Identifier::from_str_unchecked(ident),
+            Token::Crate => Identifier::from_str_unchecked(CRATE_STR),
+        };
+
+        let path = first_segment
             .then_ignore(just(Token::DoubleColon))
-            .repeated()
-            .at_least(2)
-            .collect::<Vec<_>>();
+            .then(
+                Identifier::parser()
+                    .then_ignore(just(Token::DoubleColon))
+                    .repeated()
+                    .at_least(1)
+                    .collect::<Vec<_>>(),
+            )
+            .map(|(first, mut rest)| {
+                let mut path = vec![first];
+                path.append(&mut rest);
+                path
+            });
 
         let aliased_item =
             SymbolName::parser().then(just(Token::As).ignore_then(SymbolName::parser()).or_not());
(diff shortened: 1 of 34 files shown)


apoelstra:
In c5f33a7e836427c5545990c39252d79dabca1a38:

This is full of unrelated lockfile changes.

src/test_utils.rs
```diff
@@ -2,6 +2,10 @@ use std::fs;
 use std::path::PathBuf;
 use std::time;
 
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+static WORKSPACE_COUNTER: AtomicUsize = AtomicUsize::new(0);
+
```
LesterEvSe, line 15: Explain why we need this and what problem it solves?

src/resolution.rs
```diff
@@ -192,14 +192,48 @@ impl DependencyMap {
...
+
+                    if let (Some(curr), Some(res)) = (current_crate, resolved_crate) {
+                        if curr.target == res.target {
+                            return Err(Error::LocalFileImportedAsExternal(
+                                resolved.as_path().to_path_buf(),
+                            ))
+                            .with_span(*use_decl.span());
+                        }
+                    }
+                }
```
LesterEvSe: Create a separate function for this.

src/parse.rs
```diff
@@ -1464,11 +1464,25 @@ impl ChumskyParse for UseDecl {
         // Parse the base path prefix (e.g., `dependency_root_path::file::` or `dependency_root_path::dir::file::`).
         // We require at least 2 segments here because a valid import needs a minimum
         // of 3 items total: the dependency_root_path, the file, and the specific item/function.
-        let path = Identifier::parser()
+        let first_segment = select! {
+            Token::Ident(ident) => Identifier::from_str_unchecked(ident),
+            Token::Crate => Identifier::from_str_unchecked("crate"),
```
LesterEvSe: I think it would be better to add a constant with "crate" name and use it.

LesterEvSe (changes requested):
Update the section regarding the `crate` keyword in the `doc/architecture.md` file.

src/test_utils.rs
Sdoba16: > Explain why we need this and what problem it solves?

When I run cargo test, I get a different number of errors. Sometimes I get 1 error, sometimes 2 errors, or no errors at all. This is happening because of parallel testing; different tests share the same temporary dirs (which are deleted after one test run, causing an error for the second test). So, to fix this, a counter was added which ensures that tests do not share the same temporary dir.

LesterEvSe: Add this as a docstring to the code

src/parse.rs
```diff
@@ -1464,11 +1464,25 @@ impl ChumskyParse for UseDecl {
         // Parse the base path prefix (e.g., `dependency_root_path::file::` or `dependency_root_path::dir::file::`).
         // We require at least 2 segments here because a valid import needs a minimum
         // of 3 items total: the dependency_root_path, the file, and the specific item/function.
```
KyrylR: Should we mention crate keyword here?

KyrylR:
Running https://github.com/Sdoba16/SimplicityHL/tree/feature/crate-keyword/functional-tests/valid-test-cases/external-library-uses-crate using simc directly fails in my case 

Is this expected?

src/parse.rs
LesterEvSe: Yes, I think so

Sdoba16:
> Running https://github.com/Sdoba16/SimplicityHL/tree/feature/crate-keyword/functional-tests/valid-test-cases/external-library-uses-crate using simc directly fails in my case
> 
> Is this expected?

Should work with the updates

KyrylR:
> > Running [Sdoba16/SimplicityHL@`feature`/crate-keyword/functional-tests/valid-test-cases/external-library-uses-crate](https://github.com/Sdoba16/SimplicityHL/tree/feature/crate-keyword/functional-tests/valid-test-cases/external-library-uses-crate) using simc directly fails in my case
> > Is this expected?
> 
> Should work with the updates

Could we add a regression test for this? Something like below
```rust
use std::path::{Path, PathBuf};
use std::process::Command;

fn repo_path(path: &str) -> PathBuf {
    Path::new(env!("CARGO_MANIFEST_DIR")).join(path)
}

#[test]
fn cli_dependency_can_use_crate_root() {
    let root = repo_path("functional-tests/valid-test-cases/external-library-uses-crate");
    let main = root.join("main.simf");
    let ext_lib = root.join("ext_lib");
    let dep_arg = format!("{}:ext_lib={}", root.display(), ext_lib.display());

    let output = Command::new(env!("CARGO_BIN_EXE_simc"))
        .arg(main)
        .arg("--dep")
        .arg(dep_arg)
        .output()
        .expect("failed to run simc");

    assert!(
        output.status.success(),
        "simc failed\nstatus: {:?}\nstdout:\n{}\nstderr:\n{}",
        output.status.code(),
        String::from_utf8_lossy(&output.stdout),
        String::from_utf8_lossy(&output.stderr),
    );
}
```

LesterEvSe (approved):
ACK 8f002f3fe9dd78e524f3bfbd84c4f6ae64a1cfbd; tested locally

KyrylR (approved):
ACK 8f002f3fe9dd78e524f3bfbd84c4f6ae64a1cfbd; code review, successfully ran tests locally

PR #306, opened 2026-04-29
lsp: update SimplicityHL compiler to 0.6.0-rc.0

gerau:
This is necessary changes to support new interface of compiler.

### DIFF
--- a/lsp/Cargo.toml
+++ b/lsp/Cargo.toml
@@ -21,7 +21,7 @@ thiserror = "2.0.17"
 
 ropey = "1.6.1"
 miniscript = "12"
-simplicityhl = "0.5.0-rc.0"
+simplicityhl = "0.6.0-rc.0"
 nom = "8.0.0"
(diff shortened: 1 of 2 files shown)


KyrylR:
Not sure how unresolved imports are expected to impact hover and other LSP features, but from what I can see this path now returns early when `driver::Program::from_parse` reports an `UnknownLibrary` error for a `use` declaration. Since that returns `doc = None`, the document is no longer indexed, so semantic tokens, hover, references, and completion stop working for otherwise parseable files that contain imports. Should we still create the LSP `Document` from the parsed program here, and report unresolved imports separately?

gerau:
Yep, there is no support for modules in LSP right now. I'm working on this right now so probably I will build on top of this PR, so converting it to a draft for now.

gerau:
closing because the LSP project has migrated to https://github.com/BlockstreamResearch/simplicityhl-lsp.

PR #312, opened 2026-05-05
Fix of external SDK crate override bug

### LINKED ISSUE #310: Bug: Users can override the `crate` keyword as a DRP name
### Project version

master (unreleased)

### Project

driver

### What happened?

Currently, because users can interact directly with the `DependencyMap` API in SimplicityHL, they can manually set the `crate` keyword to any path they want. This happens because the `insert` method allows adding the `crate` alias without throwing any validation errors.

## Fix Suggestion

There are two potential ways to solve this problem:
1. Add validation checks inside the `insert` method to explicitly forbid using the `crate` keyword.
2. Make the `insert` method inside the `DependencyMap` struct private to external crates. Instead, implement the `From<Vec<Remapping>>` trait to handle conversions and perform all necessary validations at that stage.

### Minimal reproduction steps

Using SDK, if you insert a mapping like the one below, it will be accepted by the compiler without any errors:

``` json
DependencyMap {
    inner: [
        Remapping {
            context_prefix: CanonPath(
                "/home/user/SimplicityHL/program",
            ),
            drp_name: "crate",
            target: CanonPath(
                "/home/user/another-dir/",
            ),
        },
    ],
}
```

Sdoba16:
Closes #310 
Previously, users could interact directly with the DependencyMap API in SimplicityHL, they could manually set the crate keyword to any path they want. This happened because the insert method allowed adding the crate alias without throwing any validation errors.

Fix
Was introduced `DependencyMapBuilder ` which collects dependencies and then could with build command it can be built.

While building are made following validations: 
- entry source must be canonical and non-anonymous
- every dependency target must be a canonical directory
- every context prefix must be a canonical directory
- alias must be a valid dependency identifier
- alias must not be keyword
- duplicate context+alias remappings rejected
- overlapping crate roots are explicitly modeled
- resolved files must remain inside their package root

Remapping was made pub(crate) to hide from external API.

`resolution.rs` was separated into 2 files `source.rs` and `resolution.rs` for readability.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -15,7 +15,7 @@ use simplicity::elements;
 
 use crate::lexer::Token;
 use crate::parse::MatchPattern;
-use crate::resolution::SourceFile;
+use crate::source::SourceFile;
 use crate::str::{AliasName, FunctionName, Identifier, JetName, ModuleName, WitnessName};
 use crate::types::{ResolvedType, UIntType};
 
@@ -474,6 +474,11 @@ impl fmt::Display for ErrorCollector {
 /// Records _what_ happened but not where.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub enum Error {
+    DependencyPathNotFound(String),
+    DependencyNotADirectory(String),
+    ReservedDependencyKeyword(String),
+    DuplicateDependencyAlias(String, String),
+    InvalidDependencyIdentifier(String),
     Internal(String),
     UnknownLibrary(String),
     ArraySizeNonZero(usize),
@@ -533,6 +538,11 @@ pub enum Error {
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
+            Error::DependencyPathNotFound(path) => write!(f, "Path not found: {}", path),
+            Error::DependencyNotADirectory(path) => write!(f, "Path must be a directory: {}", path),
+            Error::ReservedDependencyKeyword(kw) => write!(f, "The '{}' keyword is reserved and cannot be manually mapped. Use the builder's context definitions instead.", kw),
+            Error::DuplicateDependencyAlias(alias, context) => write!(f, "Duplicate dependency mapping: alias '{}' is defined multiple times for context '{}'", alias, context),
+            Error::InvalidDependencyIdentifier(alias) => write!(f, "Invalid dependency alias '{}': must be a valid identifier and not a reserved keyword", alias),
             Error::Internal(err) => write!(
                 f,
                 "INTERNAL ERROR: {err}"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,8 @@ pub mod num;
 pub mod parse;
 pub mod pattern;
 pub mod resolution;
+pub mod source;
+
 #[cfg(feature = "serde")]
 mod serde;
 pub mod str;
@@ -39,7 +41,8 @@ use crate::debug::DebugSymbols;
 use crate::driver::DependencyGraph;
 use crate::error::{ErrorCollector, WithContent, WithSource as _};
 use crate::parse::ParseFromStrWithErrors;
-use crate::resolution::{DependencyMap, SourceFile};
+use crate::resolution::DependencyMap;
+use crate::source::SourceFile;
 pub use crate::types::ResolvedType;
 pub use crate::value::Value;
 pub use crate::witness::{Arguments, Parameters, WitnessTypes, WitnessValues};
@@ -72,22 +75,17 @@ impl TemplateProgram {
                 .ok_or_else(|| error_handler.to_string())?;
 
         // 2. Create the driver program
-        let driver_program: driver::Program = if dependency_map.is_empty() {
-            driver::Program::from_parse(&parsed_program, source.content(), &mut error_handler)
-                .ok_or_else(|| error_handler.to_string())?
-        } else {
-            let graph = DependencyGraph::new(
-                source.clone(),
-                Arc::from(dependency_map.clone()),
-                &parsed_program,
-                &mut error_handler,
-            )?
-            .ok_or_else(|| error_handler.to_string())?;
+        let graph = DependencyGraph::new(
+            source.clone(),
+            Arc::from(dependency_map.clone()),
+            &parsed_program,
+            &mut error_handler,
+        )?
+        .ok_or_else(|| error_handler.to_string())?;
 
-            graph
-                .linearize_and_build(&mut error_handler)?
-                .ok_or_else(|| error_handler.to_string())?
-        };
+        let driver_program: driver::Program = graph
+            .linearize_and_build(&mut error_handler)?
+            .ok_or_else(|| error_handler.to_string())?;
 
         // 3. AST Analysis
         let ast_program = ast::Program::analyze(&driver_program).with_source(source.clone())?;
@@ -399,6 +397,10 @@ pub trait ArbitraryOfType: Sized {
 #[cfg(test)]
 pub(crate) mod tests {
     use crate::parse::ParseFromStr;
+    use crate::resolution::tests::canon;
+    use crate::resolution::DependencyMapBuilder;
+    use crate::source::CanonPath;
+    use crate::test_utils::TempWorkspace;
     use base64::display::Base64Display;
     use base64::engine::general_purpose::STANDARD;
     use simplicity::BitMachine;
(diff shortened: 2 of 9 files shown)


apoelstra:
In 3390c42baf9ff0e0ebad2c6955ae897c4cb755a2:

I'm confused by this solution. We now how have two methods, `DependencyMap::insert` and `DependencyMap::add_crate_root` which differ only in that one is private and one does a check for `"crate"`. Weirdly, the private one does the check but the public one doesn't.

In parallel to this, this commit adds `TryFrom<Vec<Remapping>> for DependencyMap` which is insufficiently general (it should be `FromIterator` rather than `TryFrom<Vec>` IMO, given that all it does is iterate over its input and call `self.insert`) and *does* do the check.

We then change `main` to construct a `Vec` which we then call `TryFrom` on, instead of just directly constructing the `DependencyMap`. What does this gain us?

KyrylR:
@Sdoba16 could you update the description to reflect recent updates?

apoelstra:
Can you squash these commits?

src/source.rs
```diff
@@ -0,0 +1,106 @@
+use std::path::Path;
```
LesterEvSe, line 1: Once this file is added, let's move the `CanonSourceFile` struct and its implementation from `driver/mod.rs` into this new one

src/lib.rs
```diff
@@ -494,33 +497,22 @@ pub(crate) mod tests {
             I: IntoIterator<Item = (P, K, P)>,
             K: Into<String>,
         {
-            let mut dependency_map = DependencyMap::new();
+            let mut builder = crate::resolution::DependencyMapBuilder::new();
```
LesterEvSe: Use `DependencyMapBuilder` instead of `crate::resolution::DependencyMapBuilder`

src/lib.rs
```diff
@@ -740,14 +731,10 @@ pub(crate) mod tests {
...
-        let mut dependency_map = DependencyMap::new();
-        let canon_root = CanonPath::canonicalize(&root).unwrap();
-        dependency_map
-            .insert(
-                canon_root.clone(),
-                crate::driver::CRATE_STR.to_string(),
-                canon_root,
-            )
+        let canon_root = crate::source::CanonPath::canonicalize(&root).unwrap();
+        let dependency_map = crate::resolution::DependencyMapBuilder::new()
```
LesterEvSe: Same here: use the name directly and remove the `crate::` prefix

src/lib.rs
```diff
@@ -1162,18 +1149,18 @@ mod error_tests {
...
     fn dependency_map(root_dir: &Path, drp: &str, lib_dir: &Path) -> DependencyMap {
-        let mut dependency_map = DependencyMap::new();
-
         let context = CanonPath::canonicalize(root_dir).unwrap();
         let target = CanonPath::canonicalize(lib_dir).unwrap();
 
-        dependency_map.insert(context, drp.into(), target).unwrap();
-
-        dependency_map
+        crate::resolution::DependencyMapBuilder::new()
```
LesterEvSe: Found `crate` again. Let’s nix these everywhere else too

src/resolution.rs
```diff
@@ -1,129 +1,154 @@
...
+/// Mappings are strictly sorted by the longest `context_prefix` match.
+/// This mathematical guarantee ensures that if multiple nested directories
+/// define the same dependency root path, the most specific (deepest) context wins.
+#[derive(Debug, Clone, Default)]
+pub struct DependencyMap {
+    /// External dependency remappings (e.g., `use math::...`)
+    remappings: Vec<Remapping>,
+    /// Package roots for resolving `crate::...` (sorted by longest path match)
+    package_roots: Vec<CanonPath>,
 }
```
LesterEvSe, line 54: Add a bit more explanation of the architecture here. Why do we need `Vec<CanonPath>` here?

src/resolution.rs
```diff
@@ -1,129 +1,154 @@
...
+impl DependencyMapBuilder {
+    pub fn new() -> Self {
+        Self::default()
     }
 
-    pub fn name(&self) -> &Option<Arc<Path>> {
-        &self.name
+    pub fn with_entry_root(mut self, root: CanonPath) -> Self {
+        self.entry_root = Some(root);
+        self
```
LesterEvSe: Let’s pull this into `new()` and get rid of the `Option` for `entry_root`

src/resolution.rs
```diff
@@ -194,61 +203,73 @@ impl DependencyMap {
...
+                    return Err(RichError::new(
+                        Error::ExternalFileNotFound(
+                            drp_name.to_string(),
+                            resolved.as_path().to_path_buf(),
+                        ),
+                        *use_decl.span(),
+                    ));
+                }
+
+                self.check_local_file_imported_as_external(current_file, &resolved, use_decl)?;
```
LesterEvSe: Extract this into a separate function

LesterEvSe (changes requested):

src/resolution.rs
```diff
@@ -1,176 +1,185 @@
...
+        crate_roots.sort();
+        crate_roots.dedup();
 
-    /// Check if the current file is executing inside the context's directory tree.
-    /// This prevents a file in `/project_a/` from using a dependency meant for `/project_b/`
-    pub fn starts_with(&self, path: &CanonPath) -> bool {
-        self.as_path().starts_with(path.as_path())
-    }
+        // Sort package roots by length descending to ensure longest prefix match
+        crate_roots.sort_by(|a, b| {
```
stringhandler, line 143: Should this not have a `then_by` sort for paths with the same length? It could probably then be used instead of the above sort at line 138

src/resolution.rs
```diff
@@ -1,176 +1,185 @@
...
-/// Mappings are strictly sorted by the longest `context_prefix` match.
-/// This mathematical guarantee ensures that if multiple nested directories
-/// define the same dependency root path, the most specific (deepest) context wins.
-#[derive(Debug, Clone, Default)]
-pub struct DependencyMap {
-    inner: Vec<Remapping>,
-}
-
 impl DependencyMap {
     pub fn new() -> Self {
```
stringhandler: Should this method maybe be private and add documentation to only use the builder. It seems like it is quite complicated to create this struct

src/resolution.rs
```diff
@@ -1,176 +1,185 @@
...
     pub fn is_empty(&self) -> bool {
-        self.inner.is_empty()
+        self.remappings.is_empty() && self.package_roots.is_empty()
     }
 
     /// Re-sort the vector in descending order so the longest context paths are always at the front.
     /// This mathematically guarantees that the first match we find is the most specific.
     fn sort_mappings(&mut self) {
-        self.inner.sort_by(|a, b| {
+        self.remappings.sort_by(|a, b| {
```
stringhandler, line 163: Also feels like there should be a second sorting clause for paths with the same length

LesterEvSe:
ACK 7ff8fd204cd6aeeb12eccc041a9bdd94496b60f7; tested locally

LesterEvSe (approved):

apoelstra (approved):
ACK 7ff8fd204cd6aeeb12eccc041a9bdd94496b60f7; successfully ran local tests

src/error.rs
```diff
@@ -474,6 +474,11 @@ impl fmt::Display for ErrorCollector {
 /// Records _what_ happened but not where.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub enum Error {
+    DependencyPathNotFound(String),
+    DependencyNotADirectory(String),
+    ReservedDependencyKeyword(String),
+    DuplicateDependencyAlias(String, String),
+    InvalidDependencyIdentifier(String),
     Internal(String),
```
KyrylR, line 482: In the context of https://github.com/BlockstreamResearch/SimplicityHL/pull/325#issuecomment-4497102830 should we reiterate on errors here? 

@gerau what do you think? (from my PoV with this PR we are adding more work for you)

src/lib.rs
```diff
@@ -72,23 +75,18 @@ impl TemplateProgram {
...
+        let graph = DependencyGraph::new(
+            source.clone(),
+            Arc::from(dependency_map.clone()),
+            &parsed_program,
+            &mut error_handler,
+        )?
+        .ok_or_else(|| error_handler.to_string())?;
+
+        let driver_program: driver::Program = graph
+            .linearize_and_build(&mut error_handler)?
```
KyrylR, line 87: Same question regarding errors here

src/error.rs
stringhandler: This PR is already ACK'ed and tested. My vote is for merging as is.

src/resolution.rs
```diff
@@ -1,176 +1,181 @@
...
-    }
+/// A router for resolving dependencies across multi-file workspaces.
+///
+/// Mappings are strictly sorted by the longest `context_prefix` match.
+/// This mathematical guarantee ensures that if multiple nested directories
+/// define the same dependency root path, the most specific (deepest) context wins.
+/// This struct must always be constructed via [`DependencyMapBuilder`].
+/// Builder guarantees that the vector is never empty and contains no duplicates, so we can safely sort without worrying about edge cases.
+#[derive(Debug, Clone, Default)]
+pub struct DependencyMap {
```
KyrylR, line 42: This still derives Default, which leaves a public way to construct an invalid map without DependencyMapBuilder

src/error.rs
gerau: I'm okay with that

KyrylR: > This PR is already ACK'ed and tested. My vote is for merging as is.

When possible I wanna avoid surprises for devs who are working on features

src/lib.rs
KyrylR: resolved by https://github.com/BlockstreamResearch/SimplicityHL/pull/312#discussion_r3279423363

src/resolution.rs
Sdoba16: Yeah, you're right, so I will just delete deriving Default, it won't break anything

KyrylR (approved):
ACK https://github.com/BlockstreamResearch/SimplicityHL/pull/312/commits/70cebfbb16f20d367b3c5d4ba850fd1650c8a1b7; 

(waiting for local tests to finish before merging)

KyrylR:
> ACK [70cebfb](https://github.com/BlockstreamResearch/SimplicityHL/commit/70cebfbb16f20d367b3c5d4ba850fd1650c8a1b7);
> 
> (waiting for local tests to finish before merging)

The diff to https://github.com/BlockstreamResearch/SimplicityHL/commit/7ff8fd204cd6aeeb12eccc041a9bdd94496b60f7 is minimal, therefore I believe there is no need for re review from others

KyrylR (approved):
ACK f2af871ceb508bd3f88cf7077c5a30334c589419; successfully ran local tests, review

Issue #224, opened 2026-03-02
Overriding the Elements jet set with the `--target` parameter


ivanlele:
### Project

compiler

### Describe the feature

I propose adding a `--target` parameter to `simc` that specifies the jet set used for SimplicityHL compilation. Based on the passed argument, the compiler would load a different [Jet trait](https://github.com/BlockstreamResearch/rust-simplicity/blob/a7d67c703f2ee180cc8f363dddb4941decdda15d/src/jet/mod.rs#L59) implementation at runtime.

Such option would facilitate a development of projects which do not use standard jets. Practical exemple is [SU](https://github.com/BlockstreamResearch/simplicity-unchained).

The implementation can be carried out in a series of PRs:
* Create JetHL trait  that defines the interfaces for all additional jet specific compilation-related computations introduced in Simplicity HL, and not available in on rust-simplicity Jet trait.
* Require `ast::Program` to accept a generic parameter for the implemented JetHL trait.
* Add a `--target` option that loads a dynamic library at runtime, whose methods would be used internally by a new CustomJetHL type.

This should be enough to make the minimal version work, although we might need to make additional changes for it to also work with codegen and LSP.

apoelstra:
concept ACK. I think we will struggle to implement this though because so many of our types are parameterized by `J: Jet` and Rust makes it very hard to swap these at runtime.

apoelstra:
@ivanlele sure, if you have a small trait then it's definitely worth PR'ing so we can discuss its merits.

What I would **like** to do is go upstream to rust-simplicity and attempt to remove the `J: Jet` parameter from the main `Node` type. It should instead live in `types::Context` (which should probably not live in `types`, but it's okay there for now), where it will prevent the user from mixing different jet types in a single program.

Then in `Node`, rather than having a generic `J: Jet`, we should have a `Box<dyn Jet>` (or maybe a `struct Jet`). However, the current `Jet` type isn't object-safe due to its associated types. So we will need to refactor the trait to invert the types, or something. Probably we should move this to a separate issue.

apoelstra:
I opened https://github.com/BlockstreamResearch/rust-simplicity/issues/349 but it's a bit rambly.

ivanlele:
Hi @apoelstra , before #314 is merged, I think we can already make progress on this issue.

My idea is to first isolate all logic related to the Jet type into a trait.

[...]

PR #314, opened 2026-05-06
chore: update rust-simplicity dependency

stringhandler:
Switch to latest  master branch of rust-simplicity, which removes the `Jet` type parameter from nodes and renames broken lock distance/ duration jets. Update all call sites to match the new API and deprecate example files that use the renamed jets.

### DIFF
--- a/src/docs/jet.rs
+++ b/src/docs/jet.rs
@@ -607,11 +607,11 @@ Using the notation of BIP-0341, it returns the SHA256 hash of c[33: 33 + 32m]."#
 - The result of [`output_surjection_proofs_hash`] (32 bytes).
 - The result of [`input_utxos_hash`] (32 bytes)."#,
         // Time locks
-        Elements::CheckLockDistance => r#"**Deprecated; do not use.** Assert that the value returned by [`tx_lock_distance`] is greater than or equal to the given value.
+        Elements::BrokenDoNotUseCheckLockDistance => r#"**Deprecated; do not use.** Assert that the value returned by [`tx_lock_distance`] is greater than or equal to the given value.
 
 ## Panics
 The assertion fails."#,
-        Elements::CheckLockDuration => r#"**Deprecated; do not use.** Assert that the value returned by [`tx_lock_duration`] is greater than or equal to the given value.
+        Elements::BrokenDoNotUseCheckLockDuration => r#"**Deprecated; do not use.** Assert that the value returned by [`tx_lock_duration`] is greater than or equal to the given value.
 
 ## Panics
 The assertion fails"#,
@@ -624,8 +624,8 @@ The assertion fails."#,
 ## Panics
 The assertion fails."#,
         Elements::TxIsFinal => "Check if the sequence numbers of all transaction inputs are at their maximum value.",
-        Elements::TxLockDistance => "**Deprecated; do not use.** If [`version`] returns 2 or greater, then return the greatest valid [`Distance`] value of any transaction input. Return zeroes otherwise.",
-        Elements::TxLockDuration => "**Deprecated; do not use.** If [`version`] returns 2 or greater, then return the greatest valid [`Duration`] value of any transaction input. Return zeroes otherwise.",
+        Elements::BrokenDoNotUseTxLockDistance => "**Deprecated; do not use.** If [`version`] returns 2 or greater, then return the greatest valid [`Distance`] value of any transaction input. Return zeroes otherwise.",
+        Elements::BrokenDoNotUseTxLockDuration => "**Deprecated; do not use.** If [`version`] returns 2 or greater, then return the greatest valid [`Duration`] value of any transaction input. Return zeroes otherwise.",
         Elements::TxLockHeight => "If [`tx_is_final`] returns false, then try to parse the transaction's lock time as a [`Height`] value. Return zeroes otherwise.",
         Elements::TxLockTime   => "If [`tx_is_final`] returns false, then try to parse the transaction's lock time as a [`Time`] value. Return zeroes otherwise.",
         // Issuance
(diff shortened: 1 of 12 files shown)


apoelstra:
@ivanlele in your opinion are we ready to just release a new version of rust-simplicity with these changes?

ivanlele:
> @ivanlele in your opinion are we ready to just release a new version of rust-simplicity with these changes?

@apoelstra, looks like it, the remaining work for #224 is only on the HL side. I was actually about to ask you what we needed to make release myself

apoelstra:
Okay, dope. @ivanlele can you open a PR on rust-simplicity which bumps the version in Cargo.toml and the lockfiles, and adds whatever CHANGELOG.md entries you think are worth mentioning? Then I can merge that, and tag and publish it.

stringhandler:
Can we possibly also get https://github.com/BlockstreamResearch/rust-simplicity/pull/362 merged before the tag?

apoelstra:
oops yes definitely. let me review it now.

ivanlele:
> Okay, dope. @ivanlele can you open a PR on rust-simplicity which bumps the version in Cargo.toml and the lockfiles, and adds whatever CHANGELOG.md entries you think are worth mentioning? Then I can merge that, and tag and publish it.

On it, I'll make it a draft for now, before Mike's PR is reviewed

ivanlele:
Bump: BlockstreamResearch/rust-simplicity#364

stringhandler:
Finally ready for review

apoelstra:
Hah, glad to see the new `broken_` names for these jets highlight where we were using them.

stringhandler:
Yeah I originally moved these examples to a new deprecated folder, but I decided to do that in a follow up PR instead

src/tracker.rs
```diff
@@ -185,11 +185,6 @@ impl<'a> DefaultTracker<'a> {
 
         let mut input_frame = input.clone();
 
-        // The reason we need to advance by a bit is that the AssertL combinator is actually a Case combinator,
-        // which takes a bit of input to decide which branch to take. But this bit is "meaningless" and
-        // is always 0 because it's an assertion.
-        let _ = input_frame.next();
```
apoelstra, line 191: In 83f6c5087b3aee61c56d22115d1f829f6961e825:

Why remove this from `handle_jet` here?

apoelstra: Similar in `parse_jet_result` below.

src/tracker.rs
```diff
@@ -317,10 +299,14 @@ impl PruneTracker<Elements> for DefaultTracker<'_> {
...
 }
 
-impl ExecTracker<Elements> for DefaultTracker<'_> {
-    fn visit_node(&mut self, node: &RedeemNode<Elements>, input: FrameIter, output: NodeOutput) {
+impl ExecTracker for DefaultTracker<'_> {
+    fn visit_node(&mut self, node: &RedeemNode, input: FrameIter, output: NodeOutput) {
         match node.inner() {
-            Inner::Jet(jet) => self.handle_jet(node, *jet, &input, &output),
+            Inner::Jet(jet) => {
+                if let Some(&elements_jet) = jet.as_any().downcast_ref::<Elements>() {
```
apoelstra, line 306: In 83f6c5087b3aee61c56d22115d1f829f6961e825:

I guess we can do this as a followup but I suspect this downcast is unnecessary and we can do everything using the generic jet.

src/tracker.rs
```diff
@@ -317,10 +299,14 @@ impl PruneTracker<Elements> for DefaultTracker<'_> {
...
+                    self.handle_jet(node, elements_jet, &input, &output);
+                }
```
apoelstra, line 310: In 83f6c5087b3aee61c56d22115d1f829f6961e825:

While we have the downcast, let's add an `else { panic! }` here so that we visibly  crash rather than silently ignoring the jet, if we are given a non-elements program.

src/tracker.rs
stringhandler: The tests fail with them. I suspect that the extra bit was actually a result of the Frame bug we fixed in rust-simplicity, and that @KyrylR (I suspect he added these) assumed it was the case that the jet had a wrapping case, when it did not.

src/tracker.rs
stringhandler: Ah yeah, let me look at this again.

src/tracker.rs
stringhandler: good idea

src/tracker.rs
ivanlele: `handle_jet` will be updated later to accept dynamic jets, so this downcast will disappear down the line.

src/tracker.rs
apoelstra: Hah, fascinating! I think you are right -- it looks like all this code comes from PRs like #198 with the attitude "let's just patch this til it works for now, and eventually we'll replace this debug-wrapping stuff completely".

src/tracker.rs
stringhandler: added

apoelstra (approved):
ACK 361966d5e39542ff9edc075b618ded6b7af31b16; successfully ran local tests

src/docs/jet.rs
```diff
@@ -607,11 +607,11 @@ Using the notation of BIP-0341, it returns the SHA256 hash of c[33: 33 + 32m]."#
 - The result of [`output_surjection_proofs_hash`] (32 bytes).
 - The result of [`input_utxos_hash`] (32 bytes)."#,
         // Time locks
-        Elements::CheckLockDistance => r#"**Deprecated; do not use.** Assert that the value returned by [`tx_lock_distance`] is greater than or equal to the given value.
+        Elements::BrokenDoNotUseCheckLockDistance => r#"**Deprecated; do not use.** Assert that the value returned by [`tx_lock_distance`] is greater than or equal to the given value.
```
KyrylR, line 610: nit: generated jet module docs still link to the old timelock jet names after the rename (i.e. to [`tx_lock_distance`], and to [`tx_lock_duration`] below)

KyrylR (approved):
ACK 361966d5e39542ff9edc075b618ded6b7af31b16; successfully ran local tests, review

PR #320, opened 2026-05-13
Feature/ci rbmt

KyrylR:
Fixes: https://github.com/BlockstreamResearch/SimplicityHL/issues/267


KyrylR:
Not sure if we need daily scheduled fuzzing (saw it here: https://github.com/ElementsProject/rust-elements/pull/267/changes)

apoelstra:
Yeah, we should add the fuzz cronjob, but let's do that in a separate PR.

apoelstra (approved):
ACK e3147ff7a10b7b8f28838f7665cc77f1a2322b2b; successfully ran local tests; this looks great!

PR #324, opened 2026-05-14
add `thiserror`, split `error::Error` into different modules

gerau:
Implementing the first step of the error refactor issue: https://github.com/BlockstreamResearch/SimplicityHL/issues/204#issuecomment-4397044768.

I tried to keep the changes minimal, so I did not refactor RichError completely. Also, keep in mind that we will most likely remove `error::Error` and replace RichError with something more like a Diagnostic.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -9,18 +9,133 @@ use simplicity::jet::{Elements, Jet};
 
 use crate::debug::{CallTracker, DebugSymbols, TrackedCallName};
 use crate::driver::{FileScoped, SymbolTable, MAIN_MODULE, MAIN_STR};
-use crate::error::{Error, RichError, Span, WithSpan};
+use crate::error::{RichError, Span, WithSpan};
 use crate::jet::JetHL;
 use crate::num::{NonZeroPow2Usize, Pow2Usize};
 use crate::parse::MatchPattern;
 use crate::pattern::Pattern;
-use crate::str::{AliasName, FunctionName, Identifier, ModuleName, WitnessName};
+use crate::str::{AliasName, FunctionName, Identifier, JetName, ModuleName, WitnessName};
 use crate::types::{
     AliasedType, ResolvedType, StructuralType, TypeConstructible, TypeDeconstructible, UIntType,
 };
 use crate::value::{UIntValue, Value};
 use crate::witness::{Parameters, WitnessTypes};
 use crate::{driver, impl_eq_hash, parse};
+
+#[derive(Debug, thiserror::Error, Clone)]
+pub enum Error {
+    #[error("Main function is required")]
+    MainRequired,
+
+    #[error("Function `{name}` was defined multiple times")]
+    FunctionRedefined { name: FunctionName },
+
+    #[error("Type alias `{name}` is not defined")]
+    UndefinedAlias { name: AliasName },
+
+    #[error("INTERNAL ERROR: {msg}")]
+    Internal { msg: String },
+
+    #[error("Item `{name}` is private")]
+    PrivateItem { name: String },
+
+    #[error("Type alias `{name}` was defined multiple times")]
+    RedefinedAlias { name: AliasName },
+
+    #[error("Expected expression of type `{expected}`, found type `{found}`")]
+    ExpressionTypeMismatch {
+        expected: ResolvedType,
+        found: ResolvedType,
+    },
+
+    #[error("Witness expressions are not allowed outside the `main` function")]
+    WitnessOutsideMain,
+
+    #[error("Witness `{name}` has been used before somewhere in the program")]
+    WitnessReused { name: WitnessName },
+
+    #[error("Function `{name}` was called but not defined")]
+    FunctionUndefined { name: FunctionName },
+
+    #[error("Failed to compile to Simplicity: {msg}")]
+    CannotCompile { msg: String },
+
+    #[error("Main function takes no input parameters")]
+    MainNoInputs,
+
+    #[error("Main function produces no output")]
+    MainNoOutput,
+
+    #[error("The 'main' function must be defined in the entry point file")]
+    MainOutOfEntryFile,
+
+    #[error("Expected expression of type `{ty}`; found something else")]
+    ExpressionUnexpectedType { ty: ResolvedType },
+
+    #[error("Variable `{identifier}` is used twice in the pattern")]
+    VariableReuseInPattern { identifier: Identifier },
+
+    #[error("Variable `{identifier}` is not defined")]
+    UndefinedVariable { identifier: Identifier },
+
+    #[error("Value is out of bounds for type `{ty}`")]
+    IntegerOutOfBounds { ty: UIntType },
+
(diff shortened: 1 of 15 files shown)


src/error.rs
```diff
@@ -475,58 +475,11 @@ impl fmt::Display for ErrorCollector {
...
-    VariableReuseInPattern(Identifier),
-    WitnessReused(WitnessName),
-    WitnessTypeMismatch(WitnessName, ResolvedType, ResolvedType),
-    WitnessReassigned(WitnessName),
-    WitnessOutsideMain,
-    ModuleRedefined(ModuleName),
-    ArgumentMissing(WitnessName),
-    ArgumentTypeMismatch(WitnessName, ResolvedType, ResolvedType),
+    LexerError(String),
+    ParsingError(crate::parse::Error),
```
stringhandler, line 479: I was expecting to see `ParsingError(#[from] crate::parse::Error)`, is there a reason it's not done this way?

src/parse.rs
```diff
@@ -31,6 +32,88 @@ use crate::str::{
 };
 use crate::types::{AliasedType, BuiltinAlias, TypeConstructible, UIntType};
 
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct SyntaxError {
```
stringhandler:
```suggestion
pub struct SyntaxErrorInfo {
```

Perhaps better to use a name that doesn't end in Error so that it is not confused with an error that can be returned

src/parse.rs
```diff
@@ -31,6 +32,88 @@ use crate::str::{
...
+    ListBoundPow2(usize),
+
+    #[error("Cannot parse: {0}")]
+    CannotParse(String),
+
+    #[error(transparent)]
+    SyntaxError(#[from] SyntaxError),
+
+    #[error("Incompatible match arms: {0}, {1}")]
+    IncompatibleMatchArms(MatchPattern, MatchPattern),
```
stringhandler:
```suggestion
    IncompatibleMatchArms{ left: MatchPattern, right: MatchPattern},
```
More of a personal preference, but I prefer using the named struct pattern when the inner members are not obvious

src/error.rs
gerau: I'm leaving the old Display implementation and not migrating to `thiserror` to keep the changes minimal. Also, I hope that we remove this enum completely, so I think it's fine to leave it as is.

src/parse.rs
gerau: Yep, would rename it

src/parse.rs
gerau: I would also like to implement this in future PRs to keep the changes minimal. The plan is to also refactor these errors to contain more information (additional spans, notes, etc.)

src/parse.rs
```diff
@@ -31,6 +32,88 @@ use crate::str::{
...
+impl Error {
+    /// Update the error with the affected span.
+    pub fn with_span(self, span: Span) -> RichError {
+        RichError::new(self.into(), span)
+    }
+}
+
+impl From<simplicity::elements::hex::Error> for Error {
+    fn from(error: simplicity::elements::hex::Error) -> Self {
+        Self::CannotParse(error.to_string())
```
apoelstra: In e656be003ce3c3baf9b439767e107678740fd23f:

In general I don't like these `From` impls -- it is better to make the user write `.map_err(some_explicit_conversion)?` rather than just `?`, because if you have the single-character `?` available, you're really discouraged from adding more error context at call sides.

Then in specific, I really don't like *these* `From` impls which convert to strings and then shove the strings into a generic enum variant.

src/ast.rs
```diff
@@ -22,6 +22,115 @@ use crate::value::{UIntValue, Value};
...
+    WitnessOutsideMain,
+
+    #[error("Witness `{0}` has been used before somewhere in the program")]
+    WitnessReused(WitnessName),
+
+    #[error("Function `{0}` was called but not defined")]
+    FunctionUndefined(FunctionName),
+
+    #[error("Failed to compile to Simplicity: {0}")]
+    CannotCompile(String),
```
apoelstra: In 39ff893bcb79d1c21559088754dde06c4830a38b:

This is used in only one place - to indicate that the `use` keyword is not yet supported. We should have a dedicated error variant for this.

src/ast.rs
```diff
@@ -22,6 +22,115 @@ use crate::value::{UIntValue, Value};
...
+    IntegerOutOfBounds(UIntType),
+
+    #[error("Cannot parse: {0}")]
+    IntParseError(String),
+
+    #[error("Expected a valid bit string length (1, 2, 4, 8, 16, 32, 64, 128, 256), found {0}")]
+    BitStringPow2(usize),
+
+    #[error("{0}")]
+    Generic(String),
```
apoelstra: In 39ff893bcb79d1c21559088754dde06c4830a38b:

What is this for? It is unused in this commit. Can it hold a `Box<dyn std::error::Error>` rather than a string?

src/ast.rs
```diff
@@ -22,6 +22,115 @@ use crate::value::{UIntValue, Value};
...
+    ArgumentMissing(WitnessName),
+
+    #[error(
+        "Parameter `{0}` was declared with type `{1}` but its assigned argument is of type `{2}`"
+    )]
+    ArgumentTypeMismatch(WitnessName, ResolvedType, ResolvedType),
+}
+
+impl From<crate::num::ParseIntError> for Error {
+    fn from(error: crate::num::ParseIntError) -> Self {
```
apoelstra: In 39ff893bcb79d1c21559088754dde06c4830a38b:

My same complaint about the `From` impls in the last commit apply here.

apoelstra:
Rather than reviewing all these commits, let me try to provide some general guidance:

* don't represent errors as `String`s unless you really have to (the two cases I know of are: if you are wrapping an error which is already a `String`; if you are holding a generic error where you have `alloc` but not `std` and a MSRV of 1.80 or less)
* don't have error messages like `cannot parse: {0}` where `{0}` is a wrapped error. Just use `cannot parse`. The user can call `source` if they want to see `{0}`
* don't use anonymous fields like `FileNotFound(String)`. Use `FileNotFound { filename: String }`. Also, in that case, consider whether the filename should actually be a `PathBuf` or a `OsString`.
* don't implement `From<WrappedError> for Error` which just stuffs the error in a wrapper, unless you are sure that there is no additional context that will ever be available

apoelstra:
I understand that you're violating many of these guidelines to keep the changes minimal so that each PR is reviewable.

Would it be possible to change only some errors at a time, but to do so completely, to create minimal PRs? For example, I think any error type which is currently a `String` could be replaced by a strongly-typed error in its own PR.

gerau:
Thanks for the guidance! 

> Would it be possible to change only some errors at a time, but to do so completely, to create minimal PRs? For example, I think any error type which is currently a String could be replaced by a strongly-typed error in its own PR.

Okay, I will try this first. Basically, I wanted to do most of this after this PR, but after your comments, I think doing it before makes more sense.

PR #325, opened 2026-05-18
simc: error: refactor generic String to error types

gerau:
For motivation see https://github.com/BlockstreamResearch/SimplicityHL/pull/324#issuecomment-4472188772.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -896,7 +896,7 @@ impl AbstractSyntaxTree for Item {
                 Function::analyze(function, ty, scope).map(Self::Function)
             }
             parse::Item::Use(use_decl) => Err(RichError::new(
-                Error::CannotCompile("The `use` keyword is not supported yet.".to_string()),
+                Error::UseKeywordIsNotSupported,
                 *use_decl.span(),
             )),
             parse::Item::Module => Ok(Self::Module),
--- a/src/error.rs
+++ b/src/error.rs
@@ -11,7 +11,6 @@ use chumsky::util::MaybeRef;
 use chumsky::DefaultExpected;
 
 use itertools::Itertools;
-use simplicity::elements;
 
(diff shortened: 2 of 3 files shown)


KyrylR:
Is it expected that let x: u8 = 999 now reports Integer parsing error, losing the previous number too large to fit in target type detail?

gerau:
I've based my changes on this comment: https://github.com/BlockstreamResearch/SimplicityHL/pull/324#issuecomment-4472185447, and if IIUC, this is pretty much expected. I would be happy to change it if you want.

KyrylR (approved):
ACK https://github.com/BlockstreamResearch/SimplicityHL/pull/325/commits/14488985ed74e29ab69e867659c7e847bf1903bb; successfully ran local tests

apoelstra:
> Is it expected that let x: u8 = 999 now reports Integer parsing error, losing the previous number too large to fit in target type detail?

Our reports should use `Error::source` to provide the full stack of error details, similar to how `anyhow` outputs errors. The other method, of manually pasting strings together in the errors' `Display` impls so that you can one-shot `fmt` the whole error stack, is (at best) inflexible since it prevents us reformatting stuff, and at worst loses information because 3rd-party errors will expect us to call `source` to get inner errors.

I also don't think "everything in `fmt`" it will survive the transition to multi-line errors.

apoelstra:
In 6a5099d2b3d9a66309ee9d26d3356fdd667356c5:

You need to impl the `source` method for `Error`.

apoelstra:
Oh, never mind, this is done in 14488985ed74e29ab69e867659c7e847bf1903bb

apoelstra (approved):
ACK 14488985ed74e29ab69e867659c7e847bf1903bb; successfully ran local tests

PR #328, opened 2026-05-25
Replace anonymous variants in the `Error` enum with named fields.

gerau:
Continuation of the errors refactoring; for motivation see https://github.com/BlockstreamResearch/SimplicityHL/pull/324#issuecomment-4472185447.

Not sure if we need to change _all_ the options of this enum. I made most of the analyzing errors in afa647f3fac2b6d1a95e775bbe79e6e36a0b3861 because they are quite obvious, so they could be dropped easily.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -728,7 +728,7 @@ impl Scope {
         let resolved_type = self
             .aliases
             .get(&global_id)
-            .ok_or_else(|| Error::UndefinedAlias(name.clone()))?;
+            .ok_or_else(|| Error::UndefinedAlias { name: name.clone() })?;
 
         // 3. Fetch the file scope for visibility checking.
         let file_scope = self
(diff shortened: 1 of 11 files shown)


apoelstra:
> Not sure if we need to change _all_ the options of this enum. I made most of the analyzing errors in [afa647f](https://github.com/BlockstreamResearch/SimplicityHL/commit/afa647f3fac2b6d1a95e775bbe79e6e36a0b3861) because they are quite obvious, so they could be dropped easily.

Agreed, but if we're unsure, I think we should err on the side of using names.

So I vote to keep the final commit. But I don't feel strongly about it at all.

apoelstra (approved):
ACK afa647f3fac2b6d1a95e775bbe79e6e36a0b3861; successfully ran local tests

apoelstra:
afa647f3fac2b6d1a95e775bbe79e6e36a0b3861 needs rebase

gerau:
rebased

KyrylR (approved):
ACK 94227613a28cf4ab1c599f867ff6007979acdda6; successfully ran local tests

apoelstra (approved):
ACK 94227613a28cf4ab1c599f867ff6007979acdda6; successfully ran local tests

PR #329, opened 2026-05-25
Add flag for unstable features

### LINKED ISSUE #317: Implement a flag for unstable features
### Project

compiler

### Describe the feature

We are currently (v0.5.0) adding a lot of new features to `simc`. Some of these features are high risk and while they may be tested, may still contain unexpected behaviour and may be changed/fixed, which would result in a different CMR being generated. 

I suggest we add an argument `-Z` or `--unstable-features` for enabling and feature flagging code that is already in the master branch so that we can:
1. Release new versions without needed to revert
2. Communicate easily with users that some features may change in future and should not be used in high risk programs.

For example, we have recently merged the `use` keyword which allows users to include multiple source files. I would suggest that this feature in particular is gated as an unstable feature. 

Out of interest, this is usually quite easy to implement, because most new features will add new keywords to the lexer and we can simply check at lexing stage whether the unstable feature is allowed or not.

Sdoba16:
Closes #317 

Was added `UnstableFeature` enum, which keeps track of unstable features.
Was added `UnstableFeatureManager` that allows to verify if the feature is presented, checks the program to have unstable features.

Was changed testing suite to have unstable feature manager. Also now if we have unstable features we make 2 runs, one with features enabled, one without them. This helps us to ensure, that there is no leakage of unstable feature, and we won't be able to use it without flag.

Now we can use `simc program.simf -Z feature` to enable unstable features while running with CLI.
For multiple features use: `simc program.simf -Z feature1 -Z feature2`

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -40,6 +40,7 @@ use crate::error::{Error, ErrorCollector, RichError, Span};
 use crate::parse::{self, ParseFromStrWithErrors};
 use crate::resolution::DependencyMap;
 use crate::source::{CanonPath, CanonSourceFile};
+use crate::UnstableFeatureManager;
 
 pub use crate::driver::resolve_order::{FileScoped, Program, SymbolTable};
 
@@ -139,6 +140,7 @@ impl DependencyGraph {
         dependency_map: Arc<DependencyMap>,
         root_program: &parse::Program,
         handler: &mut ErrorCollector,
+        unstable_manager: &UnstableFeatureManager,
     ) -> Result<Option<Self>, String> {
         let mut graph = Self {
             modules: vec![Module {
@@ -187,6 +189,7 @@ impl DependencyGraph {
                 &importer_source,
                 handler,
                 &mut queue,
+                unstable_manager,
             );
         }
 
@@ -204,6 +207,7 @@ impl DependencyGraph {
         importer_source: &CanonSourceFile,
         span: Span,
         handler: &mut ErrorCollector,
+        unstable_manager: &UnstableFeatureManager,
     ) -> Option<Module> {
         let Ok(content) = std::fs::read_to_string(path.as_path()) else {
             let err = RichError::new(
@@ -222,6 +226,10 @@ impl DependencyGraph {
         let source = CanonSourceFile::new(path.clone(), Arc::from(content));
 
         let ast = parse::Program::parse_from_str_with_errors(source.clone(), &mut error_handler);
+
+        if let Some(ref p) = ast {
+            unstable_manager.check_program(p, &mut error_handler);
+        }
 
         if error_handler.has_errors() {
             handler.extend_with_handler(source, &error_handler);
@@ -260,6 +268,7 @@ impl DependencyGraph {
 
     /// PHASE 2 OF GRAPH CONSTRUCTION: Loads, parses, and registers new dependencies.
     /// Note: This is a specialized helper function designed exclusively for the `DependencyGraph::new()` constructor.
+    #[allow(clippy::too_many_arguments)]
     fn load_and_parse_dependencies(
         &mut self,
         curr_id: usize,
@@ -268,6 +277,7 @@ impl DependencyGraph {
         importer_source: &CanonSourceFile,
         handler: &mut ErrorCollector,
         queue: &mut VecDeque<usize>,
+        unstable_manager: &UnstableFeatureManager,
     ) {
         for (path, import_span) in valid_imports {
             if inalid_imports.contains(&path) {
@@ -282,9 +292,13 @@ impl DependencyGraph {
                 continue;
             }
 
-            let Some(module) =
-                Self::parse_and_get_program(&path, importer_source, import_span, handler)
-            else {
+            let Some(module) = Self::parse_and_get_program(
+                &path,
+                importer_source,
+                import_span,
+                handler,
+                unstable_manager,
+            ) else {
                 inalid_imports.push(path);
                 continue;
             };
@@ -307,6 +321,7 @@ pub(crate) mod tests {
     use crate::resolution::tests::canon;
     use crate::resolution::DependencyMapBuilder;
     use crate::test_utils::TempWorkspace;
+    use crate::unstable::UnstableFeature;
 
     /// Initializes a raw graph environment for testing, explicitly allowing for and capturing failure states.
     ///
@@ -330,6 +345,7 @@ pub(crate) mod tests {
     /// 4. `ErrorCollector`: The handler containing any logged errors, useful fo
     pub(crate) fn setup_graph_raw(
         files: Vec<(&str, &str)>,
+        features: impl IntoIterator<Item = UnstableFeature>,
     ) -> (
         Option<DependencyGraph>,
         HashMap<String, usize>,
@@ -366,16 +382,31 @@ pub(crate) mod tests {
 
         let root_p = root_file_path.expect("main.simf must be defined in file list");
         let main_canon_source = CanonSourceFile::new(root_p, Arc::from(root_content));
+        let main_source = main_canon_source.clone();
+
+        let unstable_manager = UnstableFeatureManager::new(features);
 
         let main_program_option =
-            parse::Program::parse_from_str_with_errors(main_canon_source.clone(), &mut handler);
+            parse::Program::parse_from_str_with_errors(main_source.clone(), &mut handler);
+
+        if let Some(ref p) = main_program_option {
+            let mut unstable_errors = ErrorCollector::new();
+            unstable_manager.check_program(p, &mut unstable_errors);
+            handler.extend_with_handler(main_source.clone(), &unstable_errors);
+        }
 
         let Some(main_program) = main_program_option else {
             return (None, HashMap::new(), ws, handler);
         };
 
-        let graph_option =
-            DependencyGraph::new(main_canon_source, map, &main_program, &mut handler).unwrap();
+        let graph_option = DependencyGraph::new(
+            main_canon_source,
+            map,
+            &main_program,
+            &mut handler,
+            &unstable_manager,
+        )
+        .unwrap();
 
         let mut file_ids = HashMap::new();
 
@@ -418,8 +449,9 @@ pub(crate) mod tests {
     /// to standard error if the parser or graph builder encounters any issues.
     pub(crate) fn setup_graph(
         files: Vec<(&str, &str)>,
+        features: impl IntoIterator<Item = UnstableFeature>,
     ) -> (DependencyGraph, HashMap<String, usize>, TempWorkspace) {
-        let (graph_option, file_ids, ws, handler) = setup_graph_raw(files);
+        let (graph_option, file_ids, ws, handler) = setup_graph_raw(files, features);
 
         let Some(graph) = graph_option else {
             panic!(
@@ -437,10 +469,17 @@ pub(crate) mod tests {
         // root.simf -> "use lib::math::some_func;"
         // libs/lib/math.simf -> ""
 
-        let (graph, ids, _ws) = setup_graph(vec![
-            ("main.simf", "use lib::math::some_func;"),
-            ("libs/lib/math.simf", ""),
-        ]);
+        let (graph, ids, _ws) = setup_graph(
+            vec![
+                ("main.simf", "use lib::math::some_func;"),
+                ("libs/lib/math.simf", ""),
+            ],
+            [
+                UnstableFeature::UseKeyword,
+                UnstableFeature::CrateKeyword,
+                UnstableFeature::AsKeyword,
+            ],
+        );
 
         assert_eq!(graph.modules.len(), 2, "Should have Root and Math module");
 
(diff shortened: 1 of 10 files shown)


stringhandler:
I realise this is still in draft, so feel free to ignore this feedback.

Is it not better to have unstable features be opt in? Keep in mind that developers may already be using a lot of these constructors and you are breaking that interface.

Maybe something like this:
```rust
CompiledProgram::new(...<existing args>).with_unstable(unstable_flags);
```

Also, do we really need 3 different unstable flags for the `use`,`crate` and `as`. Surely we cannot use `crate` if `use` is not enabled? How about `UnstableFlags::Imports` as a single flag?

Sdoba16:
> Is it not better to have unstable features be opt in? Keep in mind that developers may already be using a lot of these constructors and you are breaking that interface.
> 
> Maybe something like this:
> 
> ```rust
> CompiledProgram::new(...<existing args>).with_unstable(unstable_flags);
> ```

I guess not exactly like this, because `new` will executed before `with_unstable`, and `with_unstable` will not change much. But I guess I can add something like this:
```rust
pub fn new_with_dep(
        source: CanonSourceFile,
        dependency_map: &DependencyMap,
    ) -> Result<Self, String> {
        Self::with_unstable_and_dep(source, dependency_map, &UnstableFeatureManager::default())
    }
    
 pub fn with_unstable_and_dep(
        source: CanonSourceFile,
        dependency_map: &DependencyMap,
        unstable_manager: &UnstableFeatureManager,
    ) -> Result<Self, String>   
```

It will save the existing already interface and allow to use unstable features with new constructor

Sdoba16:
> Also, do we really need 3 different unstable flags for the `use`,`crate` and `as`. Surely we cannot use `crate` if `use` is not enabled? How about `UnstableFlags::Imports` as a single flag?

That is a great point! I initially considered using a single `Imports` flag as well, but I realized both approaches have their trade-offs.

Having 3 separated provides more granularity. For example we decide to make `use` stable. Keeping 'crate' and 'as' ensures we won't accidentally make them stable at the same time. It is like a guard from mistakingly adding unstable sub-features. Also this 3 flags are better in testing of the implementation  of unstable flag. (I guess I can leave it as it is for now to help reviewers to make own tests, and before merging I can change it).

On the other hand when we use 1 flag it is much easier and we do not need to rewrite some code, if everything is stable. 

Personally I prefer security more, but if everyone will commit to second approach I will make it 1 flag only.

src/unstable.rs
```diff
@@ -0,0 +1,342 @@
...
+            "use-keyword" => Ok(UnstableFeature::UseKeyword),
+            "crate-keyword" => Ok(UnstableFeature::CrateKeyword),
+            "as-keyword" => Ok(UnstableFeature::AsKeyword),
+            _ => Err(format!("Unknown unstable feature: '{}'", s)),
+        }
+    }
+}
+
+impl UnstableFeature {
+    /// Description of what the feature does.
```
LesterEvSe: This is a redundant docstring. Let's delete it or explain more broadly why we need this functionality

src/unstable.rs
```diff
@@ -0,0 +1,342 @@
...
+    pub fn description(&self) -> &'static str {
+        match self {
+            Self::UseKeyword => "The 'use' keyword for using dependencies",
+            Self::CrateKeyword => "The 'crate' keyword for referencing the local project root",
+            Self::AsKeyword => "The 'as' keyword for aliasing imports",
+        }
+    }
+
+    /// Returns all available unstable features.
```
LesterEvSe: Same docstring note here

src/unstable.rs
```diff
@@ -0,0 +1,342 @@
...
+
+impl UnstableFeatureManager {
+    /// Creates a manager with specific features enabled.
+    pub fn new(features: impl IntoIterator<Item = UnstableFeature>) -> Self {
+        Self {
+            enabled_features: features.into_iter().collect(),
+        }
+    }
+
+    /// Checks if a specific feature is enabled.
```
LesterEvSe: Same as above regarding the docstring. Check the others accordingly

src/unstable.rs
```diff
@@ -0,0 +1,342 @@
...
+        if !self.is_enabled(feature) {
+            return Err(format!(
+                "The '{feature}' feature is not enabled.\nEnable it with: -Z {feature}"
+            ));
+        }
+        Ok(())
+    }
+
+    /// Validates that all unstable features used in a parsed program are enabled.
+    pub fn check_program(&self, program: &Program, handler: &mut ErrorCollector) {
```
LesterEvSe, line 144: I think we can simplify it like this:
``` Rust
pub fn check_program(&self, program: &Program, handler: &mut ErrorCollector) {
    for item in program.items() {
        let Item::Use(use_decl) = item else {
            continue;
        };

        let mut check_and_report = |feature| {
            if let Err(msg) = self.check_feature(feature) {
                handler.push(RichError::new(
                    Error::UnstableFeature(msg),
                    *use_decl.span(),
                ));
            }
        };

        check_and_report(UnstableFeature::UseKeyword);

        if use_decl.drp_name().is_ok_and(|drp| drp == crate::driver::CRATE_STR) {
            check_and_report(UnstableFeature::CrateKeyword);
        }

        let has_alias = match use_decl.items() {
            UseItems::Single((_, alias)) => alias.is_some(),
            UseItems::List(items) => items.iter().any(|(_, alias)| alias.is_some()),
        };

        if has_alias {
            check_and_report(UnstableFeature::AsKeyword);
        }
    }
}
```

src/unstable.rs
```diff
@@ -0,0 +1,342 @@
...
+    ) -> Result<Self, String> {
+        let mut features = HashSet::new();
+
+        for name in names {
+            let feature_name = name.as_ref().trim();
+            if feature_name.is_empty() {
+                continue;
+            }
+            features.insert(feature_name.parse::<UnstableFeature>()?);
+        }
```
LesterEvSe, line 164: You can simplify this with:
``` Rust
...
if !feature_name.is_empty() {
    features.insert(feature_name.parse::<UnstableFeature>()?);
}
...
```

src/unstable.rs
```diff
@@ -0,0 +1,342 @@
...
+    #[test]
+    fn test_from_feature_names_unknown() {
+        let result = UnstableFeatureManager::from_feature_names(vec!["unknown-feature"]);
+        assert!(result.is_err());
+        assert!(result.unwrap_err().contains("Unknown"));
+    }
+
+    // Clone tests
+    #[test]
+    fn test_manager_clone() {
```
LesterEvSe: We do not need to test that the Rust compiler's derive macros work in `test_manager_clone`, `test_manager_equality`, and `test_manager_inequality`

src/unstable.rs
```diff
@@ -0,0 +1,342 @@
...
+        assert!(manager.is_enabled(feature));
+    }
+
+    #[test]
+    fn test_new_empty() {
+        let manager = UnstableFeatureManager::new(Vec::<UnstableFeature>::new());
+        for feature in UnstableFeature::all() {
+            assert!(!manager.is_enabled(*feature));
+        }
+    }
```
LesterEvSe: `test_empty_manager` and `test_default_manager` have the exact same body. We can drop one

src/unstable.rs
```diff
@@ -0,0 +1,342 @@
...
+    #[test]
+    fn test_check_feature_error_message() {
+        let manager = UnstableFeatureManager::default();
+        let Some(&feature) = UnstableFeature::all().first() else {
+            return;
+        };
+        let error = manager.check_feature(feature).unwrap_err();
+        assert!(error.contains("not enabled"));
+        assert!(error.contains("-Z"));
+    }
```
LesterEvSe, line 251: These tests have the exact same error path. We can just combine the `assert!(error.contains(...))` checks into a single test

src/driver/mod.rs
```diff
@@ -255,6 +263,7 @@ impl DependencyGraph {
 
     /// PHASE 2 OF GRAPH CONSTRUCTION: Loads, parses, and registers new dependencies.
     /// Note: This is a specialized helper function designed exclusively for the `DependencyGraph::new()` constructor.
+    #[allow(clippy::too_many_arguments)]
```
LesterEvSe, line 271: I think it is a bad idea to have a function with 8 arguments. But anyway, I plan to refactor it during the Module implementation

LesterEvSe (changes requested):

KyrylR:
Is there a reason why CLI does not have `--unstable-features`?

KyrylR:
Needs rebase

apoelstra:
35c05d77d639ff6c7ae77e07a1852b9199ffda11 needs rebase

Sdoba16:
> Is there a reason why CLI does not have `--unstable-features`?

Now I've added it, but as `--unstable-feature`(because we enable just one feature with this flag and can use it multiple times).

LesterEvSe (approved):
ACK 446c4ed4a1e24fac1975b494a971fe62adf8d2f1; tested locally

KyrylR:
Maybe it is better to have the check on the lexer stage, or even have a dedicated preprocessing step?

Sdoba16:
> Maybe it is better to have the check on the lexer stage, or even have a dedicated preprocessing step?

I'm not sure if that's better approach.  If we only had features depending on isolated tokens, then yes. But what if we use same token for different scenarios. Here is an example: in Rust `as` keyword is used both for type casting and aliases. So if one of this features is unstable in lexer we will forbid both, or we need to check previous tokens, find `use` and check if it alias or type casting. In parser level this can be done easier due to the tree structure build on this level. 

In my opinion parser level is more flexible in different scenarios, even though it is more massive

KyrylR:
> > Maybe it is better to have the check on the lexer stage, or even have a dedicated preprocessing step?
> 
> I'm not sure if that's better approach. If we only had features depending on isolated tokens, then yes. But what if we use same token for different scenarios. Here is an example: in Rust `as` keyword is used both for type casting and aliases. So if one of this features is unstable in lexer we will forbid both, or we need to check previous tokens, find `use` and check if it alias or type casting. In parser level this can be done easier due to the tree structure build on this level.
> 
> In my opinion parser level is more flexible in different scenarios, even though it is more massive

What about dedicated preprocessing step instead?

Sdoba16:
> What about dedicated preprocessing step instead?

This actually doesn't seem like the best approach either. To implement such an idea, we would need to write our own parser. At first glance, it might seem like a small task of just finding patterns in the token list/tree. However, features can be nested, and context matters. For example, if Item1{Item2, Item3} is forbidden but Item1{Item2{Item3}} is allowed, we have to correctly track scopes and brackets. This requires building a full tree, just like an AST. If we do that, we are repeating most of the main parser's code, which means we'd have to maintain and synchronize changes across two different parsers

KyrylR:
from my PoV this implementation is too "low level", instead of treating as, crate, use keywords as unstable we could have something like 

```
enum UnstableFeature {
    Imports,
}
```

So, overall, a cleaner architecture would be:

```
struct CompilerOptions {
    unstable: FeatureSet,
}

struct FeatureUse {
    feature: UnstableFeature,
    span: Span,
}

impl UseDecl {
    fn required_unstable_features(&self, out: &mut Vec<FeatureUse>) {
        out.push(FeatureUse {
            feature: UnstableFeature::Imports,
            span: *self.span(),
        });
    }
}
```


Then the manager does not “discover” syntax by scanning arbitrary AST items feature-by-feature. AST nodes declare what they require, and the manager only checks whether those requirements are enabled. 

Then it is easier to reason about those IMO

Sdoba16:
> from my PoV this implementation is too "low level", instead of treating as, crate, use keywords as unstable we could have something like
> 
> ```
> enum UnstableFeature {
>     Imports,
> }
> ```
> 
> So, overall, a cleaner architecture would be:
> 
> ```
> struct CompilerOptions {
>     unstable: FeatureSet,
> }
> 
> struct FeatureUse {
>     feature: UnstableFeature,
>     span: Span,
> }
> 
> impl UseDecl {
>     fn required_unstable_features(&self, out: &mut Vec<FeatureUse>) {
>         out.push(FeatureUse {
>             feature: UnstableFeature::Imports,
>             span: *self.span(),
>         });
>     }
> }
> ```
> 
> Then the manager does not “discover” syntax by scanning arbitrary AST items feature-by-feature. AST nodes declare what they require, and the manager only checks whether those requirements are enabled.
> 
> Then it is easier to reason about those IMO

Yeah, I actually discussed a similar approach with @LesterEvSe recently, but using traits. The core idea is the same, but using traits helps reduce a lot of the traversal boilerplate.

For now, we decided that I should research it a bit more, so I'll put together a description and a PoC for this approach soon. I'll ping you as soon as it's ready

LesterEvSe:
Kyrylo and I discussed this further, so this might be a much better approach to implement. We can combine all the features into one big feature, such as Imports, and implement it without adding all these features to the test cases. With a native check at the parsing stage, you can see the raw code here: https://github.com/BlockstreamResearch/SimplicityHL/tree/feature/unstable-ref-with-trait

Sdoba16:
Was decided to replace this PR with #354 with better architectural decision to use distributed more high-level unstable feature managing.

PR #330, opened 2026-05-25
Reduce dependence on Elements type

ivanlele:
This PR moves forward #224. It introduces `JetHinter`, which parses jet names and provides hints on which jet to use for assertions. It comes into play at the highest possible level — main.rs — and the downstream execution flow is fully derived from it.

As a follow-up, I plan to implement `JetHL` for `Core` with an appropriate `JetHinter` build function and add tests so we can see the effect even better.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -1,12 +1,11 @@
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::num::NonZeroUsize;
-use std::str::FromStr;
 use std::sync::Arc;
 
 use either::Either;
 use miniscript::iter::{Tree, TreeLike};
-use simplicity::jet::Elements;
+use simplicity::jet::{Elements, Jet};
 
 use crate::debug::{CallTracker, DebugSymbols, TrackedCallName};
 use crate::driver::{FileScoped, SymbolTable, MAIN_MODULE, MAIN_STR};
@@ -271,10 +270,11 @@ impl Call {
 impl_eq_hash!(Call; name, args);
 
 /// Name of a called function.
-#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+#[derive(Clone, Debug, Eq, Hash)]
+#[allow(clippy::derived_hash_with_manual_eq)] // see comment on manual `PartialEq` impl below
 pub enum CallName {
-    /// Elements jet.
-    Jet(Elements),
+    /// Jet type.
+    Jet(Box<dyn JetHL>),
     /// [`Either::unwrap_left`].
     UnwrapLeft(ResolvedType),
     /// [`Either::unwrap_right`].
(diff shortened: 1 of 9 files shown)


src/ast.rs
```diff
@@ -515,14 +539,38 @@ impl TreeLike for ExprTree<'_> {
     }
 }
 
+type JetParser = fn(&str) -> Option<Box<dyn JetHL>>;
+
+#[derive(Debug, Clone)]
+pub struct JetHinter {
```
apoelstra: In cf7273ba29f3f651b9806d39d22df6a70eb8a831:

Heh, this struct is sorta like a manual implementation of C++ inheritance. You've got an explicit vtable in the form of a function pointer and then you're holding the "concrete type" in a `Box<dyn>` which has its own vtable.

I think this would make more sense, or at least be more Rustic, if it were

```rust
/// Object which produces a specific kind of jet.
///
/// All methods return a `dyn Jet` rather than the specific jet so that the trait itself
/// can be object-safe. However, implementors of this trait **must** ensure that
/// all methods return the same kind of jet to avoid panics.
///
/// Users may rely on this property for correctness of their code, though since this
/// is a safe trait, of course they may not rely on it for soundness.
pub trait JetHinter {
    /// Attempts to parse a jet from a string.
    fn parse_jet(&str) -> Result<Box<dyn Jet>, Box<dyn std::error::Error>>;
    /// Constructs an instance of the `verify` jet.
    fn construct_verify() -> Box<dyn Jet>;
}

struct ElementsJetHinter {
    // ...
}

impl ElementsJetHinter {
    fn new() -> Self { /* ... */ }
}
```

(I added `construct_` to `verify()` because it was too easy to read `verify` as a verb, like "verify the `JetHinter`", rather than a noun, as "construct a `jet::verify` instance".)

Then everywhere you use the `JetHinter` struct you can use a `Box<dyn JetHinter>` or `&dyn JetHinter`.

src/jet.rs
```diff
@@ -3,16 +3,51 @@ use crate::types::BuiltinAlias::*;
 use crate::types::UIntType::*;
 use crate::types::*;
 
+use simplicity::jet::DynJet;
 use simplicity::jet::Elements;
 use simplicity::jet::Jet;
 
-pub trait JetHL {
+pub trait JetHLDyn {
```
apoelstra: In cf7273ba29f3f651b9806d39d22df6a70eb8a831:

I'm unclear on the purpose of this trait. It seems like the given methods can just be put directly onto `JetHL`. (And the `clone_box` method can just be replaced with a `Clone` bound on `JetHL`.)

apoelstra (approved):
ACK 6e0d6c1df19e6305e50d7df17771056877a4933f; successfully ran local tests

PR #331, opened 2026-05-25
ci: add daily fuzz job

gerau:
Addresses https://github.com/BlockstreamResearch/SimplicityHL/pull/320#issuecomment-4471497046. 

Mostly copied from rust-bitcoin and from this PR: https://github.com/ElementsProject/rust-elements/pull/267.


.github/workflows/cron-daily-fuzz.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          persist-credentials: false
+      - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        id: cache-fuzz
+        with:
+          path: |
+            ~/.cargo/bin
+            fuzz/target
+            target
+          key: cache-${{ matrix.fuzz_target }}-${{ hashFiles('**/Cargo.toml','**/Cargo.lock') }}
```
gerau, line 42: I also changed this line since it seems to be a bug -- we only have `matrix.fuzz_target`, but in the rust-bitcoin it references `matrix.target`

original: https://github.com/rust-bitcoin/rust-bitcoin/blob/9e21bc479bb0041af64a9d50b76672f82597981c/.github/workflows/cron-daily-fuzz.yml#L136

apoelstra: Neat! I'll raise an issue on rust-bitcoin. It may be that we're failing to cache build artifacts in the fuzz job because of this.

apoelstra (approved):
ACK d23cc59eb40aa54fcb63aede666ff71d3e2644a0; successfully ran local tests

PR #332, opened 2026-05-26
test: add witness pruning test verifying serialized output size

stringhandler:
Creates a test to check that witnesses are pruned along with programs.

# Motivation
I was looking into how pruned programs are checked on elements, and I saw that there are rules that the witness cannot have any extra data. The problem is that witness nodes may be removed from the program tree during pruning. 

This test proves that the witness is removed when calling `to_vec_with_witness`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1001,6 +1001,7 @@ pub(crate) mod tests {
             ),
             Arguments::default(),
             false,
+            Box::new(ElementsJetHinter::new()),
         )
         .expect("program compiles");
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -982,6 +982,65 @@ pub(crate) mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


apoelstra:
It would also be good to check that witnesses are "shrunk", i.e. if we take a single `u32` witness and only look at the MSB, the witness should be shrunk to a single bit.

This might not be possible in SimplicityHL, which uses jets for all arithmetic, since jets always take the whole value. But maybe we could take a `u16`, destructure it into a `(u8, u8)` and then read only one `u8`?

canndrew:
@stringhandler: I opened a PR against this PR to add @apoelstra's test case and fix the formatting.

apoelstra:
@canndrew can you gpg-sign your commits? Then @stringhandler can you directly point your branch to his commits, rather than merging, so that we avoid merge commits in the PR and we also preserve all signatures?

Alternately you can just copy the commits, sign them yourself, and give Andrew a coauthor credit.

PR #333, opened 2026-05-27
just: add local check_all

KyrylR:
Added `check_fuzz_crate`, `check_fuzz_bins`, and `check_all` in [`justfile`](justfile), locked integration build/test recipes (kept the full `check_fuzz` fuzz-run out of `check_all`) updated the integration helper for the current `JetHinter` API in [`test.rs`](bitcoind-tests/tests/common/test.rs), refreshed [`Cargo.lock`](bitcoind-tests/Cargo.lock), and removed the unused `test_mempool_accept` helper from [`daemon.rs`](bitcoind-tests/tests/common/daemon.rs)

### DIFF
--- a/justfile
+++ b/justfile
@@ -44,13 +44,43 @@ check_fuzz:
 build_fuzz:
     cargo-fuzz check
 
+# Check the standalone fuzz crate compiles
+check_fuzz_crate:
+    cargo rbmt --lock-file existing run --toolchain nightly -- check --manifest-path fuzz/Cargo.toml
+
+# Run fuzz target unit tests
+check_fuzz_bins:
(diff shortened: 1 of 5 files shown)


justfile
```diff
@@ -44,13 +44,43 @@ check_fuzz:
 build_fuzz:
     cargo-fuzz check
 
+# Check the standalone fuzz crate compiles
+check_fuzz_crate:
+    cargo rbmt --lock-file existing run --toolchain nightly -- check --manifest-path fuzz/Cargo.toml
```
delta1, line 49: This doesn't work in the `nix develop` shell, but I'll fix that up in a follow-up

delta1 (approved):
utACK 9e4061a0f7e5ed33173db8149f7a8de40c179a7f

PR #334, opened 2026-05-27
Implement Core JetHinter

ivanlele:
This PR introduces the CoreJetHinter implementation, demonstrating the changeability of the main executable jet set.

### DIFF
--- a/src/jet/core.rs
+++ b/src/jet/core.rs
@@ -1,0 +1,815 @@
+use crate::jet::JetHL;
+use crate::types::BuiltinAlias::*;
+use crate::types::UIntType::*;
+
+use super::*;
+
+use simplicity::jet::{Core, Jet};
+
+impl JetHL for Core {
+    fn source_jet_classification(&self) -> SourceJetClassification {
+        source_jet_classification(*self)
+    }
+
+    fn target_jet_classification(&self) -> TargetJetClassification {
+        target_jet_classification(*self)
+    }
+
+    fn is_disabled(&self) -> bool {
+        matches!(self, Core::CheckSigVerify | Core::Verify)
+    }
+
+    fn clone_box(&self) -> Box<dyn JetHL> {
+        Box::new(*self)
+    }
+
+    fn as_jet(&self) -> &dyn Jet {
+        self
+    }
+}
+
+fn source_jet_classification(jet: Core) -> SourceJetClassification {
+    match jet {
+        /*
+         * ==============================
+         *          Core jets
+         * ==============================
+         *
+         * Multi-bit logic
+         */
+        Core::Low1
+        | Core::Low8
+        | Core::Low16
+        | Core::Low32
+        | Core::Low64
+        | Core::High1
+        | Core::High8
+        | Core::High16
+        | Core::High32
+        | Core::High64 => SourceJetClassification::Unary,
+        Core::Verify => SourceJetClassification::Custom(vec![bool()]),
+        Core::Complement1
+        | Core::Some1
+        | Core::LeftPadLow1_8
+        | Core::LeftPadLow1_16
+        | Core::LeftPadLow1_32
+        | Core::LeftPadLow1_64
+        | Core::LeftPadHigh1_8
+        | Core::LeftPadHigh1_16
+        | Core::LeftPadHigh1_32
+        | Core::LeftPadHigh1_64
+        | Core::LeftExtend1_8
+        | Core::LeftExtend1_16
+        | Core::LeftExtend1_32
+        | Core::LeftExtend1_64
+        | Core::RightPadLow1_8
+        | Core::RightPadLow1_16
+        | Core::RightPadLow1_32
+        | Core::RightPadLow1_64
+        | Core::RightPadHigh1_8
+        | Core::RightPadHigh1_16
+        | Core::RightPadHigh1_32
+        | Core::RightPadHigh1_64 => SourceJetClassification::Unary,
+        Core::Complement8
+        | Core::Some8
+        | Core::All8
+        | Core::Leftmost8_1
+        | Core::Leftmost8_2
+        | Core::Leftmost8_4
+        | Core::Rightmost8_1
+        | Core::Rightmost8_2
+        | Core::Rightmost8_4
+        | Core::LeftPadLow8_16
+        | Core::LeftPadLow8_32
+        | Core::LeftPadLow8_64
+        | Core::LeftPadHigh8_16
+        | Core::LeftPadHigh8_32
+        | Core::LeftPadHigh8_64
+        | Core::LeftExtend8_16
+        | Core::LeftExtend8_32
+        | Core::LeftExtend8_64
+        | Core::RightPadLow8_16
+        | Core::RightPadLow8_32
+        | Core::RightPadLow8_64
+        | Core::RightPadHigh8_16
+        | Core::RightPadHigh8_32
+        | Core::RightPadHigh8_64
+        | Core::RightExtend8_16
+        | Core::RightExtend8_32
+        | Core::RightExtend8_64 => SourceJetClassification::Unary,
+        Core::Complement16
+        | Core::Some16
+        | Core::All16
+        | Core::Leftmost16_1
+        | Core::Leftmost16_2
+        | Core::Leftmost16_4
+        | Core::Leftmost16_8
+        | Core::Rightmost16_1
+        | Core::Rightmost16_2
+        | Core::Rightmost16_4
+        | Core::Rightmost16_8
+        | Core::LeftPadLow16_32
+        | Core::LeftPadLow16_64
+        | Core::LeftPadHigh16_32
+        | Core::LeftPadHigh16_64
+        | Core::LeftExtend16_32
+        | Core::LeftExtend16_64
+        | Core::RightPadLow16_32
+        | Core::RightPadLow16_64
+        | Core::RightPadHigh16_32
+        | Core::RightPadHigh16_64
+        | Core::RightExtend16_32
+        | Core::RightExtend16_64 => SourceJetClassification::Unary,
+        Core::Complement32
+        | Core::Some32
+        | Core::All32
+        | Core::Leftmost32_1
+        | Core::Leftmost32_2
+        | Core::Leftmost32_4
+        | Core::Leftmost32_8
+        | Core::Leftmost32_16
+        | Core::Rightmost32_1
+        | Core::Rightmost32_2
+        | Core::Rightmost32_4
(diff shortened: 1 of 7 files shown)


ivanlele:
@apoelstra After this PR, we can finally do the thing. We can load external implementations for Jet, JetHL, and JetHinter using dlopen2. I'll try to set up a minimal working example with it and a few jets.

src/jet/core.rs
```diff
@@ -0,0 +1,816 @@
...
+        Box::new(*self)
+    }
+
+    fn as_jet(&self) -> &dyn Jet {
+        self
+    }
+}
+
+pub fn source_type(jet: Core) -> Vec<AliasedType> {
+    match jet {
```
apoelstra, line 32: In bf0a1d89f1092336102c5015754724dd64267829:

Can you replace this massive match with something that parses the [`TypeName` returned from `Jet::source_ty` and `Jet::target_ty`](https://docs.rs/simplicity-lang/latest/simplicity/jet/trait.Jet.html#tymethod.source_ty)?

ivanlele: I've dreamt of doing this for months. Right now, this function returns several aliases instead of native types, example: Pubkey, Message64, Scalar, and Gej. If we could drop these aliased types here in favor of parsing them into native types, that would be great.

The thing is, that this might break something during syntax analysis, but i'm not sure.

Ideally, it would be cool to resolve these aliases in some preprocessing step so they become native types everywhere in the codebase before being fed to the lexer. But, I'm not sure whether that would negatively affect the quality of errors outputed by the compiler, which Chumsky currently handles well.

Pros and cons I guess.

ivanlele: >  Right now, this function returns several aliases instead of native types

Just to point out -- this is the main reason why we can't just parse it to target_ty, because parsing can't be deterministic, I forgot to highlight it

apoelstra: Can we `match` and only manually handle the cases that have `Pubkey`, `Message64`, etc.? And then use a `_` wildcard match for all other jets?

This would address all the arithmetic jets, which are the vast majority of them.

apoelstra: I see, even with the arithmetic jets we have things like `vec![U8.into(), U8.into()]` for jets whose input signature actually looks like `U16`. But conceptually these are "binary jets" that take two inputs.

ivanlele: We can. on it

apoelstra: I would suggest we do something like

```rust
enum JetClassification {
    Simple,
    BinaryInput,
    Custom(Vec<AliasedType>),
}
```

(this should be a private enum). Then we can implement source_type like

```rust
fn source_type(jet: Core) -> Vec<AliasedType> {
    match jet_classification(jet) {
        JetClassification::Simple => parse_ty(jet.source_ty()),
        JetClassification::BinaryInput => split_in_half(parse_ty(jet.source_ty())),
        JetClassification::Custom(vec) => vec,
    }
}
```

where `jet_classification` has a giant match in it, but ideally this is the only function with a giant match. And it's not trying to write out full type signatures, it's just doing a loose classification.

Does this make sense?

apoelstra: Basically, I really want there to be a "single source of truth" for these types. I understand that for the crypto jets where we have "rich types" for inputs and outputs, this is not really possible. But there aren't a ton of these jets.

ivanlele: I think it deserves a separate PR

ivanlele: While experimenting with this today, I found that it's not that easy to determine which number types should be used in different cases. For example, with a notation like `**lll`, it's not deterministic whether it should represent `[u64, u64, u64]`, `[u128, u64]`, or some other composition. Similarly, a notation such as `****22*22**22*22***22*22**22*22` gives us the same: should it be interpreted as `[u8, u8]` or as `[u16]` or something else?

A potential solution would be to introduce a stricter notation on the generated `source_ty` and `target_ty` side. We already have the following spec:

```rust
/// | char | type         |
/// |------|--------------|
/// | `1`  | unit         |
/// | `2`  | single bit   |
/// | `c`  | 8-bit word   |
/// | `s`  | 16-bit word  |
/// | `i`  | 32-bit word  |
/// | `l`  | 64-bit word  |
/// | `h`  | 256-bit word |
```

We could change `*_ty` methods to avoid uncertainty. Instead of notations like `****22*22**22*22***22*22**22*22` which is `[u8, u8]` or `[u16]`, we could encode them more directly, for example, `*cc` would represent `[u8, u8]`, while `s` would represent `[u16]`.

For `u128`, we could introduce a new symbol, for example `m`, to represent a 128-bit word.

ivanlele:
> @apoelstra After this PR, we can finally do the thing. We can load external implementations for Jet, JetHL, and JetHinter using dlopen2. I'll try to set up a minimal working example with it and a few jets.

I've created a draft PR for this branch that implements this feature.

I initially wanted to use `dlopen2` to support multiple platforms, but it requires a Rust version higher than our MSRV. So, I wrote a small driver to do the same myself for that too:
ivanlele/SimplicityHLFork#10

src/jet/core.rs
ivanlele: @apoelstra Does this approach work for you? It would require changes to both simplicity and rust-simplicity, so perhaps we could proceed with the implementation in this PR for now and work this out right way later.

ivanlele:
> > @apoelstra After this PR, we can finally do the thing. We can load external implementations for Jet, JetHL, and JetHinter using dlopen2. I'll try to set up a minimal working example with it and a few jets.
> 
> I've created a draft PR for this branch that implements this feature.
> 
> I initially wanted to use `dlopen2` to support multiple platforms, but it requires a Rust version higher than our MSRV. So, I wrote a small driver to do the same myself for that too: [ivanlele#10](https://github.com/ivanlele/SimplicityHLFork/pull/10)

If we want this before `Core` jets support, I can rebase my draft on top of `master`. But having a `jet` directory would help to keep the project structure cleaner, as we can keep all jets related staff there

src/jet/core.rs
apoelstra: @ivanlele I don't think we can change the encoding like this. These encoding eventually come from the Haskell Rust-generation code, and that far back in the pipeline there really isn't a difference between `[u8, u8]` and `u16`. Arguably there is no difference until this crate, where we're exposing these to the user as part of a high-level language.

So in the end we **will** need a giant ad-hoc match somewhere in this crate. I just want it to be as "high level" as we can manage.

apoelstra: > it's not deterministic whether it should represent `[u64, u64, u64]`, `[u128, u64]`, or some other composition

For this case I think it's fine to just hardcode the type. In my proposal above, we'd use `JetClassification::Custom([u128, u64])` for `Core::DivMod128_64`.

We should have a `Ternary` classification so for `Median8`/`Median16`/`Median32`/`Median64` we would use `JetClassification::Ternary` which would interpret the same type as `[u64, u64; u64]`.

ivanlele: #344 addresses this

ivanlele:
@apoelstra I rebased this one after #344 was merged.

src/jet/core.rs
```diff
@@ -0,0 +1,815 @@
...
+        | Core::Subtract8
+        | Core::Multiply8
+        | Core::Le8
+        | Core::Lt8
+        | Core::Min8
+        | Core::Max8
+        | Core::DivMod8
+        | Core::Divide8
+        | Core::Modulo8
+        | Core::Divides8 => SourceJetClassification::Binary,
```
stringhandler, line 260: Sorry I didn't get to reviewing the PR where these names came in, but I find these names (`SourceJetClassification::Uniary, Binary, Tertiary, Quad..`) extremely vague and particularly ambiguous in the case of Binary.

I cannot really tell if this is correct or not without understanding the logic behind them.

After reading the logic I can see what it is, but I can't help but think that this is less helpful (and maybe more error prone) than specifying everything as `SourceJetClassification::Custom()` with the actual aliased type. 

It may even be better to replace all of this with a generated file that creates the `Custom` classification

stringhandler (changes requested):

src/jet/core.rs
ivanlele: Basically, we did this to keep information about inputs and outputs as close to `C Simplicity` as possible.

Unfortunately, we cannot auto-generate these because the resulting output would not be deterministic, as discussed in the comments here:
https://github.com/BlockstreamResearch/SimplicityHL/pull/334#pullrequestreview-4373707965

As for the naming, I think the names serve their purpose quite well.
The words `unary`, `binary`, and `ternary` literally mean "consisting of "number"" respectively:
https://dictionary.cambridge.org/dictionary/english/ternary

"Consisting of three" does a good job of acting as a hint for the parser; it tells that the type consists of three elements, so the parser needs to split it into three parts.

I can only think of "number"-part" as a synonym, but it does not sound well with the `Custom` case.

ivanlele:
Do we need something else for this PR?

src/jet/core.rs
stringhandler: I don't know if there's any need to keep closely to `C Simplicity`. In most cases we are trying to simplify the complicated interface it has. 

When I first looked at this code, I needed to implement this code for a new jet. I looked at a similar jet which said it took `SourceJetClassification::Binary`. My first thought was that it took binary data as input (not "consisting of two"), so I traced it back see what the code actually did. That is why I raised this issue. 

I don't want to halt the progress here. We can merge it as is but just wanted to highlight the confusing nature of these names, even if they appear in the cambridge dictionary.

ivanlele: I see a point where it might appear confusing; maybe a good slap of documentation can solve this. I’ll try to write a few lines to cover SourceJetClassification and TargetJetClassification.

stringhandler: No need. If you are assuming devs will read documentation then you can assume they will read the source code.

stringhandler: This discussion is outside the scope of this PR though. I'll open an issue instead

apoelstra: Yes, please open an issue. This classification I specifically requested because I didn't want to have to sift through hundreds of "custom" jet types to see if they matched the actual Simplicity type.

apoelstra (approved):
ACK 793872812eadc213e69107caa4b7b62dec72dc65; successfully ran local tests

KyrylR (approved):
ACK https://github.com/BlockstreamResearch/SimplicityHL/commit/793872812eadc213e69107caa4b7b62dec72dc65; successfully ran local tests

PR #336, opened 2026-05-28
feat: add enum declarations and multi-arm enum match expressions

stringhandler:
Introduces `enum` with explicit u8 discriminants and N-arm `match` over enum variants, desugared into `jet::eq_8` comparison chains at the AST level. Missing witness values are zero-filled before Simplicity witness population; a post-prune check errors if any zero-filled witness appears on a surviving (non-pruned) branch.

# Example
```rust

// last_will.simf


enum Action {
    Inherit=1,
    ColdSpend =2,
    HotSpend =3,
}

fn main() {
    match witness::ACTION {
        Action::Inherit => { 
            let inheritor_sig: Signature = witness::INHERITOR_SIG;
            inherit_spend(inheritor_sig)} ,
        Action::ColdSpend => {
            let cold_sig: Signature = witness::COLD_SIG;
            cold_spend(cold_sig) },
        Action::HotSpend => {
            let hot_sig: Signature = witness::HOT_SIG;
            refresh_spend(hot_sig) },
    }
}

```

```json
// last_will.inherit.wit
{
    "ACTION": {
        "value": "1",
        "type": "u8"
    },
    "INHERITOR_SIG": {
        "value": "0x755201bb62b0a8b8d18fd12fc02951ea3998ba42bfc6664daaf8a0d2298cad43cdc21358c7c82f37654275dc2fea8c858adbe97bac92828b498a5a237004db6f",
        "type": "Signature"
    }
}

```

# Implementation notes

Some things to note that are missing:
## 1. Enums with structured fields. 
I originally wanted to implement enums as something like
```rust
enum Path {
   Inherit { signature: Signature}
}
```

or potentially even with tuples, like `Inherit(<inner tuple>)`. This would certainly look cleaner in the witness file (i.e. as `ACTION: { value = "Inherit("0x....")}`, however the enum declaration is not present in scope when parsing the file and this makes decoding it difficult.
As a trade off enums must have a u8 encoding, so that they can be parsed from the witness. This explicit declaration should also help when the transaction is examined on the explorer.

Structured enums can possibly be implemented in a future PR.

## 2. Empty witness filling
By splitting the action/path witness variable into it's own `u8`, we now have to specify individual witness variables for each of the actions/paths taken in the program. This means that we would now have to specify values for all witnesses, even if they are not called.

I did a deep dive and confirmed that unused witness values are pruned, but calling a program with a witness file or programmatically is still inconvenient. To combat this, I've added a step to fill unspecified witness values with defaults of zero, and then an additional step to make sure that none of the filled values remain when pruned. This is done through `satisfy_with_env`, so lib callers should not have to do any extra lifting.

## 3. Enum values in code are out of scope

A limit of the implementation is that using enums in code is not supported. This should be done in another PR.
For example, the following is not supported:
```rust

enum Direction {
  Up =1,
  Down = 2
}

fn return_up() -> Direction {
  Direction::Up
}

```
This is unfortunate but the PR is already quite large and tricky to review.

## 4. Desugaring into `jet::eq_u8`  chains
Under the hood, the `match` statement is desugared into a list of `ifs` (actually `match` if you are specific), each with a jet::eq_u8. 
While it might have been more efficient to utilize an `Either::Left` structure, hoping for a tree like evaluation, it was difficult to decode a u8 reliably into Left/Right nodes and allow for skipping enum values (e..g `enum ACTION { One = 1, /* Two = 2 --deprecated */, Three = 3 } `). That said, it is not impossible, and I can (reluctantly)  create another PR for that implementation if desired.


apoelstra:
3666ffb0040bca632166f78ef8a71ed6d0454414 needs rebase

LesterEvSe:
0eb01e6437eb49cffc9cbe4a156162aa8e9773e8 needs rebase

LesterEvSe:
Since we added unstable feature support, I think we need to add the `enum` keyword to `UnstableFeature`. Check the `doc/unstable-features.md` and `unstable.rs` files for more information

src/parse.rs
```diff
@@ -460,6 +463,91 @@ impl_eq_hash!(TypeAlias; name, ty);
...
+impl AsRef<Span> for EnumVariant {
+    fn as_ref(&self) -> &Span {
+        &self.span
+    }
+}
+
+/// An enum declaration.
+#[derive(Clone, Debug)]
+pub struct EnumDeclaration {
+    file_id: usize,
```
LesterEvSe: Currently, `file_id` is accessible via the `Span` struct, so we can delete it here. Try to do something similar to the other structs inside the `Item` struct

src/ast.rs
```diff
@@ -935,6 +978,32 @@ impl Scope {
...
+        // An enum is also a `u8` type alias, so its name resolves as a type.
+        let resolved = self.resolve(&AliasedType::from(UIntType::U8))?;
+        self.current_module_mut()
+            .aliases
+            .insert(name.clone(), (resolved, visibility));
+        self.current_module_mut()
+            .enums
+            .insert(name, EnumBinding::new(variants));
+        Ok(())
+    }
```
LesterEvSe, line 1028: Throughout the PR there's a lot of code where statements are packed together with no blank lines between logically distinct steps, which makes the individual blocks hard to scan.

Could we add blank lines to separate the logical sections? A couple of simple conventions go a long way. For example, a blank line before a `return`/`Ok(())` that closes a block, and a blank line before a comment that introduces a new step.

```suggestion
    pub fn insert_enum(
        &mut self,
        name: AliasName,
        visibility: Visibility,
        variants: Arc<[ResolvedEnumVariant]>,
    ) -> Result<(), Error> {
        if self.current_module().enums.contains_key(&name)
            || self.current_module().aliases.contains_key(&name)
        {
            return Err(Error::RedefinedAlias { name });
        }

        // An enum is also a `u8` type alias, so its name resolves as a type.
        let resolved = self.resolve(&AliasedType::from(UIntType::U8))?;
        self.current_module_mut()
            .aliases
            .insert(name.clone(), (resolved, visibility));
        self.current_module_mut()
            .enums
            .insert(name, EnumBinding::new(variants));

        Ok(())
    }
```

src/ast.rs
```diff
@@ -1134,6 +1203,53 @@ impl AbstractSyntaxTree for Item {
...
+                let n = decl.variants().len();
+                if n < 2 {
+                    return Err(Error::Grammar {
+                        msg: format!("enum '{}' must have at least 2 variants", decl.name()),
+                    })
+                    .with_span(decl);
+                }
+                let mut sorted: Vec<&parse::EnumVariant> = decl.variants().iter().collect();
+                sorted.sort_by_key(|v| v.discriminant());
+                for w in sorted.windows(2) {
```
LesterEvSe, line 1244: Same as above; blank lines between logical steps

src/ast.rs
```diff
@@ -1447,6 +1563,75 @@ impl AbstractSyntaxTree for SingleExpression {
...
+                let arms = enum_match.arms();
+                let span = *enum_match.span();
+                if arms.is_empty() {
+                    return Err(Error::Grammar {
+                        msg: "enum match has no arms".to_string(),
+                    })
+                    .with_span(span);
+                }
+                let enum_name = match arms[0].pattern() {
+                    MatchPattern::EnumVariant(name, _) => name.clone(),
```
LesterEvSe, line 1608: Blank lines between logical steps

src/ast.rs
```diff
@@ -1447,6 +1563,75 @@ impl AbstractSyntaxTree for SingleExpression {
...
+                    _ => unreachable!("EnumMatch arms have EnumVariant patterns"),
```
LesterEvSe, line 1609: `unreachable!` will panic the whole process if it's ever hit, and this runs on caller-supplied witnesses/env. The mismatch may be impossible today, but a change in pruned representation could reach it. The function already returns `Result`, so let's return an `Err` here instead and fail gracefully

src/ast.rs
```diff
@@ -1447,6 +1563,75 @@ impl AbstractSyntaxTree for SingleExpression {
...
+                let binding = scope
+                    .get_enum(&enum_name)
+                    .ok_or_else(|| Error::UndefinedAlias {
+                        name: enum_name.clone(),
+                    })
+                    .with_span(span)?;
+                let mut arm_map: HashMap<&Identifier, &parse::Expression> = HashMap::new();
+                for arm in arms {
+                    let MatchPattern::EnumVariant(arm_enum_name, variant) = arm.pattern() else {
+                        unreachable!("EnumMatch arms have EnumVariant patterns")
```
LesterEvSe, line 1622: Same comment regarding the `unreachable!` block

src/named.rs
```diff
@@ -243,6 +244,67 @@ pub fn populate_witnesses(
...
+        }
+        // Case pruned to AssertR: only right branch survived
+        (Inner::Case(_, cr), Inner::AssertR(_, pr)) => {
+            check_surviving_witnesses(cr, pr, zero_filled)
+        }
+        // Disconnect — not used in SimplicityHL; handle defensively
+        (Inner::Disconnect(cc, _), Inner::Disconnect(cp, _)) => {
+            check_surviving_witnesses(cc, cp, zero_filled)
+        }
+        _ => unreachable!("unexpected structural mismatch between commit and pruned trees"),
```
LesterEvSe: Another `unreachable` block

LesterEvSe (changes requested):

src/named.rs
stringhandler: I have replaced this with an Err

src/ast.rs
stringhandler: This unreachable and the one below it seem reasonable to me. I don't like the idea of returning an Error here because that tells me we expect that kind of thing to happen, which it is very unlikely given that we construct the tree.

I considered other patterns but all of them are more complicated than this and I don't see the benefit

src/ast.rs
stringhandler: See above

src/ast.rs
stringhandler: I added some blank lines but I don't know what you had in mind

src/ast.rs
stringhandler: I added some but this is very subjective, so not sure what you had in mind

src/ast.rs
stringhandler: Added

src/parse.rs
stringhandler: removed

stringhandler:
Added to unstable features

LesterEvSe (approved):
ACK cfa1116d4a91eeeb4e638abd7f2df2b875833b37; tested locally with `just check` and `just check_fuzz`

LesterEvSe:
I talked with Kyrylo a bit, and he helped me find a few more crucial bugs in the code

LesterEvSe:
imported enums lose their variant metadata.
`Action` resolves as a TYPE (`u8` alias) after `use`, so `let a: Action = ...`
works, but matching `Action::A` fails because the `EnumBinding` variant list
was never copied into the importing module's scope.

Test case to reproduce:
```rust
mod m {
    pub enum Action {
        A = 0,
        B = 1,
    }
}

use crate::m::Action;

fn main() {
    let a: Action = 0;
    match a {
        Action::A => { },
        Action::B => { },
    }
}
```

So this code failed with a `Type alias Action is not defined` error

LesterEvSe:
The zero-fill safety check is skipped when `env = None`.

The program has two branches. Branch B needs witness `B`. We select branch B (SELECTOR = 2) but do not provide witness `B`, and we ask for zero-filling.

With `env = Some(..)`  -> pruning runs, check_surviving_witnesses runs, the fabricated zero `B` is detected.
With `env = None` -> pruning is skipped, so the check is skipped too, and the fabricated zero `B` slips through.

A missing live witness (e.g. a required signature) is silently forged as zero.

You can test it with adding this code to the `tests/` directory:
```rust
use simplicityhl::ast::ElementsJetHinter;
use simplicityhl::str::WitnessName;
use simplicityhl::value::ValueConstructible;
use simplicityhl::{Arguments, CompiledProgram, UnstableFeatures, Value, WitnessValues};
use std::collections::HashMap;

const SRC: &str = r#"
enum Branch { A = 1, B = 2 }
fn main() {
    match witness::SELECTOR {
        Branch::A => assert!(jet::is_zero_32(witness::A)),
        Branch::B => assert!(jet::is_zero_32(witness::B)),
    }
}
"#;

fn compile() -> CompiledProgram {
    CompiledProgram::new_with_unstable(
        SRC,
        &UnstableFeatures::all(),
        Arguments::default(),
        false,
        Box::new(ElementsJetHinter::new()),
    )
    .unwrap()
}

#[test]
fn zerofill_env_none_skips_safety_check() {
    let mut map: HashMap<WitnessName, Value> = HashMap::new();
    map.insert(WitnessName::from_str_unchecked("SELECTOR"), Value::u8(2));
    // NOTE: witness `B` is intentionally NOT provided.

    let env = simplicityhl::dummy_env::dummy();
    let with_env = compile().satisfy_with_env_fill_missing(
        WitnessValues::from(map.clone()),
        Some(&env),
        true,
    );
    assert!(
        with_env.is_err(),
        "sanity check: with env, the missing live witness B must be rejected"
    );
    println!("with env  -> Err (correct): {}", with_env.unwrap_err());

    let without_env =
        compile().satisfy_with_env_fill_missing(WitnessValues::from(map), None, true);
    println!(
        "without env -> {}",
        if without_env.is_ok() { "Ok(satisfied)" } else { "Err" }
    );

    assert!(
        without_env.is_err(),
        "BUG: without env, missing live witness B was zero-filled and accepted"
    );
}
```

LesterEvSe:
I tried to run the shipped example with `cargo run --quiet -- -Z enums examples/last_will.simf --wit examples/last_will.inherit.wit` and got `Error: "missing witness for HOT_SIG"`.

I think this is the same root cause as the `env = None` issue: the CLI still calls the strict satisfy `main.rs:225`, so it requires a value for every witness

KyrylR:
Closed in favor of #376

PR #337, opened 2026-05-29
Implement module system and program flattening

### LINKED ISSUE #318: Support `mod` keyword with Namespaces
## Project

compiler

## Motivation

The long-term goal is to run Simplicity programs in the browser via [Simplex](https://github.com/BlockstreamResearch/smplx). To support multi-file projects, they would need to be compiled into a single entry point before being passed to Simplex to eliminating filesystem dependencies and establishing a single source of truth for the program.

Our driver can already flatten multiple files into a single `Vec<Item>` and inject visibility rules into the structure. However, this is not sufficient for correct composition: the flat merge breaks visibility boundaries and alias resolution, producing an invalid program.

## Main Idea

Treat each imported file as a separate module by wrapping it in a `mod` block. this preserves namespaces and proper scoping, allowing items to be accessed safely via qualified paths:

``` Rust
let bar_res: u32 = mod_block::bar_fn();
```

## Proposed Semantics

The module system is modelled after [Rust's module system](https://doc.rust-lang.org/rust-by-example/mod.html), adapted for Simplicity. The following features are in scope:

- **Visibility:** `pub mod { /* ... */ }`
- **Nested modules:** Because one namespace can be called from the another.
- **Extended `use` declarations:** Currently `use` only allows importing specific items (e.g. `use math::crypto::sha256`). We propose extending it to also support qualified path access inside code (e.g. `module_a::bar()`) without requiring an explicit `use` import.

Waiting for comments from @apoelstra, @stringhandler, @KyrylR to discuss the concept of implementation.

LesterEvSe:
## Overview

Implements the module system proposed in #318, enabling multi-file programs to be compiled into a single entry point for browser execution via [Simplex](https://github.com/BlockstreamResearch/smplx).

## Changes

- [x] **Driver**: Each imported file is now wrapped in a `mod unit_N { ... }` block during flattening, preserving namespace boundaries and visibility rules
- [x] **Analyzer**: added `ModuleScope` and scope-aware resolution to handle nested modules, visibility checks, and `use` rewriting across module boundaries

## Supported Syntax

``` Rust
pub mod foo { ... }          // visibility
mod outer { mod inner { } }  // nesting
use crate::foo::bar;         // explicit import
```

Related PRs (in order of the merge): #338, #339, #343, #345


LesterEvSe:
**Next Steps After Merging:**
We expect the modules feature to be marked as unstable in PR #329.
After merging this and the Unstable Features PR, we plan to bump the SimplicityHL version to `0.6.0-rc.1`.

KyrylR (approved):
ACK 4359a845206a8c7f11d4bc03d6dc9674f135e060; based on the ACKs in: #338, #339, #343, #345

delta1 (approved):
ACK 4359a845206a8c7f11d4bc03d6dc9674f135e060; tested locally

LesterEvSe:
**Context for faster review.**

**What changed?**
Most of the driver logic has been moved into `ast.rs` and the module system has been introduced.
Previously, the driver handled file loading, alias resolution, visibility checks, and file ordering. Now it is responsible only for file loading, ordering, and flattening.

**Flattening**
The driver used to collect all `Items` into a flat `Vec<Item>` for `AST::analyze` to process. This has been restructured: each file is now wrapped in a `mod unit_N { ... }` block, converting a multi-file program into a single-file representation while preserving the same CMR (more details in 4359a845206a8c7f11d4bc03d6dc9674f135e060).

**Scopes in `ast.rs`**
A scope system has been introduced, modelled as a DAG data structure (see commit ab629c6b56bcca9822f47c76a8c3720ac539d6da for details). Visibility and alias resolution are now handled natively at this stage inside `Scope::resolve_use(...)`. Each `mod` block adds a new node to the DAG and creates a new scope.

KyrylR:
Ran all tests successfully again (`just check_all`)

PR #338, opened 2026-05-29
Add module parser

LesterEvSe:
Implement a working `Module` struct and handle recursive parsing. In this series of PRs, we do not plan to support qualified calls via `::` (e.g., `unit_1::call_function`), as our existing alias system is sufficient.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -74,8 +74,10 @@ pub enum Item {
     TypeAlias,
     /// A function.
     Function(Function),
-    /// A module, which is ignored.
+    Use,
     Module,
+    /// A placeholder used for error recovery during parsing.
+    Ignored,
 }
 
 /// Definition of a function.
@@ -984,7 +986,11 @@ impl AbstractSyntaxTree for Item {
                 Error::UseKeywordIsNotSupported,
                 *use_decl.span(),
             )),
-            parse::Item::Module => Ok(Self::Module),
+            parse::Item::Module(module) => Err(RichError::new(
+                Error::ModuleKeywordIsNotSupported,
+                *module.span(),
+            )),
+            parse::Item::Ignored => Ok(Self::Ignored),
         };
 
         scope.file_id = previous_file_id;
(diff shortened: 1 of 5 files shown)


src/ast.rs
```diff
@@ -74,8 +74,10 @@ pub enum Item {
     TypeAlias,
     /// A function.
     Function(Function),
-    /// A module, which is ignored.
+    Use,
     Module,
+    /// A placeholder used for error recovery during parsing.
+    Ignored,
```
KyrylR, line 80: Let's have a dedicated issue for the need of this keyword

src/parse.rs
```diff
@@ -847,6 +844,7 @@ impl TreeLike for ExprTree<'_> {
     }
 }
 
+// TODO: Fix the formatter
```
KyrylR: nit: for the future let's have more reasoning in text for the todos

src/str.rs
```diff
@@ -325,6 +319,41 @@ impl From<SymbolName> for ModuleName {
 
 wrapped_string!(ModuleName, "module name");
 
+#[cfg(feature = "arbitrary")]
+impl<'a> arbitrary::Arbitrary<'a> for ModuleName {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        // TODO: Consider to change it
```
KyrylR: nit: we should move it out of the local scope 

reserved names sound like something that other functions can benefit from

src/ast.rs
LesterEvSe: Add documentation for that

src/parse.rs
LesterEvSe: Added

src/str.rs
LesterEvSe: Moved the constants into the crate's scope and added public functions for the corresponding checks

KyrylR (approved):
ACK c835f1aea6ebc075733d1fc3b9621b0ca76e7141; successfully ran local tests

PR #342, opened 2026-06-01
simplicity-hl: introduce contract id

### LINKED ISSUE #293: Suggestion: Creating a unique identifier for programs: Canonical CMR and parameter slots
### Project

core

### Describe the feature

This concept could also be logged in `rust-simplicity` or even `simplicity` but will create it here since it deals with parameters.

# Problem 
Wallets will need a way of identifying simplicity programs for deterministic recovery. Something similar to miniscript fragments used for wallet descriptors. 

Wallets will most likely store the SimplicityHL source code and recompile them with parameters. I am suggesting an identifier for this source code.

# Canonical CMR
Since the program already has the concept of CMR's, I suggest we uniquely identify a program through this process:

1. Substitute each parameter (word in the compiled code) with an index per type. So the first u32 would be 0, second u32 would be 1, third parameter might be a u256, which would be 0. (Note: the ordering is up for debate, but I would suggest using the same post node ordering that encoding uses)
2. Generate a CMR using this code.

The program can then be encoded as `<ccmr><param_type><param_type>....`. 

This has the added benefit of allowing parameter substitution into a compiled simplicity program.

## Notes
1. Originally, I used zero for all parameter values, but this has the consequence of generating a different node structure due to node sharing. This could still be considered, but there could be multiple programs that compile to the same CCMR. Another reason is that round trip encoding of programs should remain the same. As it is, a round trip encoding/decoding of a program with duplicate parameters would result in different encodings. 
2. I suggest a unique index per type, instead of unique overall, otherwise some indexes may overflow the type. Even with this adjustment, the ccmr would be limited to 2 `boolean` args, 4 `u2` args,
3. Calculating CCMRs from onchain or pruned programs. Originally I came up with the concept of CCMRs while working on [every-simplicity](https://stringhandler.github.io/every-simplicity/). I was trying to match programs from the block explorer back to their source code. However, `case` nodes are pruned and will differ if there are parameters stored in their branches.

I am happy to hear other solutions.

KyrylR:
Fixes: https://github.com/BlockstreamResearch/SimplicityHL/issues/293

### DIFF
--- a/src/contract_id.rs
+++ b/src/contract_id.rs
@@ -1,0 +1,171 @@
+use std::collections::HashMap;
+use std::fmt;
+use std::sync::Arc;
+
+use simplicity::Cmr;
+
+use crate::error::{RichError, WithContent as _};
+use crate::types::{ResolvedType, TypeInner, UIntType};
+use crate::value::{Value, ValueConstructible};
+use crate::{named, Arguments, Parameters, TemplateProgram};
+
+/// Canonical identifier for a SimplicityHL contract template.
+///
+/// The identifier is the CMR of the template compiled with every parameter set
+/// to the zero value of its declared type and debug symbols disabled.
+#[derive(Copy, Clone, Eq, PartialEq, Hash, PartialOrd, Ord)]
+pub struct ContractId(Cmr);
+
+impl ContractId {
+    /// Compute the contract ID for the given template.
+    ///
+    /// The contract ID is the CMR of the template compiled with every parameter
+    /// set to the zero value of its declared type.
+    ///
+    /// Debug symbols are disabled during the compilation.
+    ///
+    /// ## Errors
+    ///
+    /// Returns an error if program compilation fails.
+    pub fn from_template(template: &TemplateProgram) -> Result<Self, RichError> {
+        let arguments = zero_arguments(template.parameters());
+        let commit = template
+            .simfony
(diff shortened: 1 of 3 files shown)


KyrylR:
CI failed with weird error...

image

https://github.com/BlockstreamResearch/SimplicityHL/actions/runs/26762791346/job/78880736030

KyrylR:
> CI failed with weird error...
> image
> 
> [BlockstreamResearch/SimplicityHL/actions/runs/26762791346/job/78880736030](https://github.com/BlockstreamResearch/SimplicityHL/actions/runs/26762791346/job/78880736030)

It looks like a test binary failed at link time because nightly rust-lld crashed...

KyrylR:
> > CI failed with weird error...
> > image
> > [BlockstreamResearch/SimplicityHL/actions/runs/26762791346/job/78880736030](https://github.com/BlockstreamResearch/SimplicityHL/actions/runs/26762791346/job/78880736030)
> 
> It looks like a test binary failed at link time because nightly rust-lld crashed...

Again: https://github.com/BlockstreamResearch/SimplicityHL/actions/runs/26762791346/job/78883619136?pr=342

apoelstra (changes requested):
I apologize for getting to this late, but I think we want to go a different direction with this. Will comment on the issue.

KyrylR:
Closing it for now, will sync on it on meeting

PR #343, opened 2026-06-02
Add a new strategy for import resolution and rename variables

LesterEvSe:
- Rename `Module` struct to `SourceModule` and other variables to use more appropriate and descriptive names.
- Update the loading strategy to align with the new `Module` functionality.


KyrylR (approved):
ACK 86a010098fe4ea5f1089e8952ff9679b694bfd61; successfully ran local tests

PR #354, opened 2026-06-10
feat: unstable with traits

### LINKED ISSUE #317: Implement a flag for unstable features

Sdoba16:
Closes #317

Replaces #329

Added `UnstableFeature` enum to track available unstable features
Added `UnstableFeatures` to check whether a feature is enabled

Added `RequireFeature` trait, implemented for every AST node. Each node reports which unstable features it requires. Struct impls use full field destructuring and enum impls use exhaustive matches, so new nodes or fields cause a compile error until the impl is updated.

The check runs at parse time: once a program parses without errors, any use of a disabled feature is reported with an error naming the feature and the `-Z` flag that enables it.

Added a `-Z <feature>` flag to the CLI. Multiple features can be passed with repeated flags. Available features are listed in the help output.

`simc program.simf -Z imports`

Added `doc/unstable-features.md` with user and developer guides.

### DIFF
--- a/doc/unstable-features.md
+++ b/doc/unstable-features.md
@@ -1,0 +1,17 @@
+# Unstable features
+
+Unstable features are experimental compiler capabilities. They may change
+or be removed before stabilization.
+
+## Viewing available unstable features
+
+Run `simc --help`; the features are listed under the `-Z` flag.
+
+## Enabling an unstable feature
+
+Pass `-Z <feature-name>` to `simc`.
+
+## Adding or stabilizing a feature
+
+See the rustdoc on `UnstableFeature` in `src/unstable.rs`; the procedures
+live next to the code they mutate, so they cannot drift.
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -37,6 +37,7 @@ use crate::error::{Error, ErrorCollector, RichError, Span};
 use crate::parse::{self, ParseFromStrWithErrors};
 use crate::resolution::{DependencyMap, ResolvedUse};
 use crate::source::{CanonPath, CanonSourceFile};
+use crate::unstable::UnstableFeatures;
 
 /// The reserved identifier for the program's entry point.
 pub(crate) const MAIN_STR: &str = "main";
@@ -179,6 +180,7 @@ impl DependencyGraph {
         dependency_map: Arc<DependencyMap>,
         root_program: &parse::Program,
         handler: &mut ErrorCollector,
+        unstable_features: &UnstableFeatures,
     ) -> Result<Option<Self>, String> {
         let mut graph = Self {
             modules: vec![SourceModule {
@@ -227,6 +229,7 @@ impl DependencyGraph {
                 invalid_imports: &mut invalid_imports,
                 handler,
                 queue: &mut queue,
+                unstable_features,
             };
             graph.load_and_parse_dependencies(&current, valid_imports, &mut ctx);
         }
@@ -247,6 +250,7 @@ impl DependencyGraph {
         importer_source: &CanonSourceFile,
         span: Span,
         handler: &mut ErrorCollector,
+        unstable_features: &UnstableFeatures,
     ) -> Option<SourceModule> {
         let Ok(content) = std::fs::read_to_string(path.as_path()) else {
             let err = RichError::new(
@@ -264,7 +268,11 @@ impl DependencyGraph {
         let mut error_handler = ErrorCollector::new();
         let source = CanonSourceFile::new(path.clone(), Arc::from(content));
 
-        let ast = parse::Program::parse_from_str_with_errors(source.clone(), &mut error_handler);
+        let ast = parse::Program::parse_from_str_with_errors(
+            source.clone(),
+            unstable_features,
+            &mut error_handler,
+        );
 
         if error_handler.has_errors() {
             handler.extend_with_handler(source, &error_handler);
@@ -322,9 +330,13 @@ impl DependencyGraph {
                 continue;
             }
 
-            let Some(module) =
-                Self::parse_and_get_source_module(&path, &current.source, import_span, ctx.handler)
-            else {
+            let Some(module) = Self::parse_and_get_source_module(
+                &path,
+                &current.source,
+                import_span,
+                ctx.handler,
+                ctx.unstable_features,
+            ) else {
                 // Safe to ignore output: previous `.contains` check prevents collisions.
                 let _ = ctx.invalid_imports.insert(path);
                 continue;
@@ -422,6 +434,7 @@ struct LoadContext<'a> {
     invalid_imports: &'a mut HashSet<CanonPath>,
     handler: &'a mut ErrorCollector,
     queue: &'a mut VecDeque<usize>,
+    unstable_features: &'a UnstableFeatures,
 }
 
 /// The currently processed module and its source, used for error reporting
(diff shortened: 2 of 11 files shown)


src/unstable.rs
```diff
@@ -0,0 +1,289 @@
...
+use std::str::FromStr;
+
+use crate::error::{Error, ErrorCollector, RichError, Span};
+use crate::source::SourceFile;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum UnstableFeature {
+    /// Import and module syntax, including `use`, `crate::`, and import aliases.
+    Imports,
+}
```
LesterEvSe: Let's mention that module syntax and import aliases use the `mod` and `as` keywords, respectively. Also, let's add a brief explanation about why we are combining all of these into a single block

src/driver/mod.rs
```diff
@@ -230,6 +234,10 @@ impl DependencyGraph {
 
         let ast = parse::Program::parse_from_str_with_errors(source.clone(), &mut error_handler);
 
+        if let Some(program) = &ast {
+            unstable_features.check_program(program, &source.clone().into(), &mut error_handler);
+        }
+
```
LesterEvSe: I think we can move this feature check inside the `parse_from_str` and `parse_from_str_with_errors` methods, right after the AST is successfully built. It feels like this validation should be encapsulated within the parsing module, rather than making the driver responsible for orchestrating it

doc/unstable-features.md
```diff
@@ -0,0 +1,120 @@
...
+
+- *Enum impls* use exhaustive `match self` — adding a new variant is a compile error until the
+  new arm is written.
+- *Struct impls* use full destructuring (`let Struct { field_a, field_b: _, ... } = self`) —
+  adding a new field is a compile error until the destructuring pattern is updated.
+- `UnstableFeature::all()` contains a `_check_exhaustive` const fn that exhaustively matches
+  every variant, ensuring the returned slice stays in sync with the enum.
+
+`UnstableFeatures` stores enabled features in a `Vec`. A linear scan is fast enough given the
+small number of features that will ever exist.
```
LesterEvSe: Let's move these implementation details closer to the code. Put these explanations directly on the specific methods and structs instead of keeping them in the doc

doc/unstable-features.md
```diff
@@ -0,0 +1,120 @@
...
+   exhaustive `match self` in the parent's impl will fail to compile — add the delegate call there.
+5. Add tests that compile successfully with the feature enabled and fail with a clear error when
+   the feature is disabled.
+
+## Developer guide: Stabilizing an unstable feature
+
+Remove the feature variant from `UnstableFeature`. The compiler will report errors in `Display`,
+`FromStr`, `description()`, `all()`, and every `FeatureRequirement::new(...)` call that references
+the removed variant — follow the errors to clean up. Tests that are not specifically about feature
+gating should use the helpers that enable all features, so they continue to pass after stabilization.
```
LesterEvSe: I saw that we use `#[allow(dead_code)]` in the tests inside `lib.rs`. Should we add a sentence here to remind developers to clean those up during stabilization?

LesterEvSe (changes requested):

src/unstable.rs
```diff
@@ -0,0 +1,433 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct UnstableFeatures {
+    enabled_features: Vec<UnstableFeature>,
+}
+
+impl UnstableFeatures {
+    pub fn none() -> Self {
+        Self::default()
+    }
```
LesterEvSe:
``` suggestion
    #[inline]
    pub fn none() -> Self {
        Self::default()
    }
```

src/unstable.rs
```diff
@@ -0,0 +1,433 @@
...
+        let mut uses = Vec::new();
+        program.feature_requirements(&mut uses);
+        for req in uses {
+            if !self.is_enabled(req.feature) {
+                let error = Error::UnstableFeature {
+                    feature: req.feature,
+                };
+                handler.push(RichError::new(error, req.span).with_source(source.clone()));
+            }
+        }
```
LesterEvSe:
``` suggestion
    let mut uses = Vec::new();
    program.feature_requirements(&mut uses);

    for FeatureRequirement { feature, span } in uses {
        if self.is_enabled(feature) {
            continue;
        }
        let error = Error::UnstableFeature { feature };
        handler.push(RichError::new(error, span).with_source(source.clone()));
    }
```

src/unstable.rs
```diff
@@ -0,0 +1,433 @@
...
+                } else {
+                    Some(s.parse::<UnstableFeature>())
+                }
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+        // Normalize like `new` so the sorted + deduped invariant holds regardless
+        // of constructor (e.g. `-Z imports -Z imports` collapses to one entry),
+        // reusing the `Vec` we already collected for the `?` short-circuit.
+        Ok(Self::from_sorted_dedup(features))
+    }
```
LesterEvSe:
``` suggestion
    pub fn from_names(
        names: impl IntoIterator<Item = impl AsRef<str>>,
    ) -> Result<Self, String> {
        let features = names
            .into_iter()
            .filter_map(|n| {
                let s = n.as_ref().trim();
                (!s.is_empty()).then(|| s.parse::<UnstableFeature>())
            })
            .collect::<Result<Vec<_>, _>>()?;

        // Normalize like `new` so the sorted + deduped invariant holds regardless
        // of constructor (e.g. `-Z imports -Z imports` collapses to one entry),
        // reusing the `Vec` we already collected for the `?` short-circuit.
        Ok(Self::from_sorted_dedup(features))
    }
```

src/parse.rs
```diff
@@ -1157,15 +1314,17 @@ impl<A: ChumskyParse + std::fmt::Debug> ParseFromStr for A {
     }
 }
 
-impl<A: ChumskyParse + std::fmt::Debug> ParseFromStrWithErrors for A {
+impl<A: ChumskyParse + RequireFeature + std::fmt::Debug> ParseFromStrWithErrors for A {
     fn parse_from_str_with_errors(
         source: impl Into<SourceFile>,
+        unstable_features: &UnstableFeatures,
         handler: &mut ErrorCollector,
     ) -> Option<Self> {
```
LesterEvSe, line 1275: Let's rewrite the function with this body:

``` suggestion
        let source: SourceFile = source.into();
        let src = source.content().to_string();

        let (tokens, lex_errs) = crate::lexer::lex(&src);
        let lex_ok = lex_errs.is_empty();
        handler.extend(source.clone(), lex_errs);
        let tokens = tokens?;

        let eoi = (src.len()..src.len()).into();
        let (ast, parse_errs) = A::parser()
            .parse(tokens.as_slice().map(eoi, |(t, s)| (t, s)))
            .into_output_errors();
        let parse_ok = parse_errs.is_empty();
        handler.extend(source.clone(), parse_errs);

        if let (Some(ast), true) = (&ast, lex_ok && parse_ok) {
            unstable_features.check_program(ast, &source, handler);
        }

        // TODO: We should return parsed result if we found errors, but because analyzing in `ast` module
        // is not handling poisoned tree right now, we don't return parsed result
        if handler.has_errors() {
            None
        } else {
            ast
        }
```

LesterEvSe (changes requested):

doc/unstable-features.md
Sdoba16: Added that to the Stabilizing section: there's now a paragraph reminding devs to drop the `#[allow(dead_code)]` attributes

LesterEvSe (approved):
ACK dbe633b602c946fd0f927aeed31c6045c6e88785; tested locally

LesterEvSe:
Note that once this PR is merged, we will need to update [Simplex](https://github.com/BlockstreamResearch/smplx/tree/dev) to support the new unstable features.

apoelstra:
131cc53c5d8df02aaa6090a6a0f837a23a6ca294 needs rebase

KyrylR (approved):
ACK b8d2b343fce4da412a7ab66daf31657b6014de4f; successfully ran local tests

PR #355, opened 2026-06-11
tracker: use generic JetHL instead of hardcoding Elements

KyrylR:
Refactor tracker to support changes in #334

Now, when `CoreJetHinter` is passed to `prune_with_tracker`, the function will not panic with the message that `Elements Jet` was expected.

If an unexpected Jet is encountered, it will be piped to the warning sink

### DIFF
--- a/src/tracker.rs
+++ b/src/tracker.rs
@@ -1,12 +1,12 @@
 use simplicity::bit_machine::{ExecTracker, FrameIter, NodeOutput, PruneTracker, SetTracker};
-use simplicity::jet::{Elements, Jet};
 use simplicity::node::Inner;
 use simplicity::{Ihr, RedeemNode, Value as SimValue};
 
 use crate::array::Unfolder;
+use crate::ast::{ElementsJetHinter, JetHinter};
 use crate::debug::{DebugSymbols, TrackedCallName};
 use crate::either::Either;
-use crate::jet::{source_type, target_type};
+use crate::jet::{source_type, target_type, JetHL};
 use crate::str::AliasName;
 use crate::types::AliasedType;
 use crate::value::StructuralValue;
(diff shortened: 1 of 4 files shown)


src/tracker.rs
```diff
@@ -303,10 +315,10 @@ impl ExecTracker for DefaultTracker<'_> {
     fn visit_node(&mut self, node: &RedeemNode, input: FrameIter, output: NodeOutput) {
         match node.inner() {
             Inner::Jet(jet) => {
-                if let Some(&elements_jet) = jet.as_any().downcast_ref::<Elements>() {
-                    self.handle_jet(node, elements_jet, &input, &output);
+                if let Some(jet) = self.jet_hinter.conjure(jet.as_ref()) {
+                    self.handle_jet(node, jet.as_ref(), &input, &output);
                 } else {
-                    panic!("Expected an Elements jet, found a different type of jet: {jet:?}");
```
KyrylR, line 309: @stringhandler not sure whether there was a particular reason to have a panic here

If anything, let me know if it is ok to move it to warn

ivanlele: Basically, it was added when we pulled changes from `rust-simplicity` that included dynamic dispatch. Maybe we can trigger `expect` on cast in JetHinter `conjure` implementation. The best approach would be to make `ExecTracker::visit_node` return an error, but this method is from `rust-simplicity`, so we're kinda stuck here :(

ivanlele:
ACK 147f040ecc55a4f59939a09121d7ab9b6ed37bee; ran tests locally

apoelstra (approved):
ACK 147f040ecc55a4f59939a09121d7ab9b6ed37bee; successfully ran local tests

PR #356, opened 2026-06-11
Optimize `DependencyMapBuilder` for multiple entry roots

### LINKED ISSUE #353: Refactor(DependencyMapBuilder): support reuse across multiple entry roots with shared dependencies
## Motivation

When managing imports in [Simplex](https://github.com/BlockstreamResearch/smplx), the compiler needs to build several entry points that share the same dependency graph but differ only in their `entry_root`.

## Problem

`DependencyMapBuilder` currently requires an `entry_root` in `new()`, and calling `build()` consumes the struct. This means the entire builder must be reconstructed from scratch for each entry root, re-validating all dependencies every time — even though they have not changed.

## Proposed Solution

Separate dependency validation from entry root resolution into two distinct phases:

Phase 1: Validate and freeze dependencies: `O(n log n)`, runs once
Phase 2: Attach a new entry root: `O(n)`, runs per entry point

This may involve introducing an intermediate struct (e.g. `ValidatedDeps`) that holds the pre-validated, pre-sorted dependencies and exposes a cheap `with_root(entry_root)` method to produce a `DependencyMap` without re-validating anything.

LesterEvSe:
Introduce a new `ValidatedDeps` structure and remove `entry_root` from `DependencyMapBuilder`. This also optimizes the build algorithm complexity from `O(n²)` to `O(n log n)`.

Finally, this refactors the tests to accommodate the new functionality and adds helper macros to make testing more convenient.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -1059,7 +1059,6 @@ impl Program {
         })
     }
 
-    // Put this helper function inside your impl block or just above your logic
     fn extract_single_main(items: &[Item]) -> Result<Option<Expression>, Error> {
         let mut main_expr = None;
 
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -397,8 +397,7 @@ struct CurrentModule {
 #[cfg(test)]
 pub(crate) mod tests {
     use super::*;
-    use crate::resolution::tests::canon;
-    use crate::resolution::DependencyMapBuilder;
+    use crate::resolution::tests::{build_map, canon};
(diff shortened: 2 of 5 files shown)


apoelstra:
IMHO these macros obfuscate the code more than they help.

LesterEvSe:
OK, I will change it to a function

KyrylR (approved):
ACK b31c413fe97543f3143fc88b79d391a27cba5a1c; successfully ran local tests

PR #357, opened 2026-06-12
External jets loading

ivanlele:
This PR introduces a new type of jet that can be loaded through a shared library and used to compile Simplicity programs outside the current scope.

It has been tested on Linux, macOS, and Windows.

Progress #224.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -106,3 +106,22 @@ jobs:
             echo "fuzzing $target"
             cargo fuzz run --fuzz-dir fuzz "$target" -- -runs=1
           done
+
+  ExternalJetLib:
+    name: External Jet Library
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout repo"
+        uses: actions/checkout@v4
+
+      - name: "Setup cargo-rbmt"
+        uses: ./.github/actions/setup-rbmt
+
+      - name: "Build external jet dylib"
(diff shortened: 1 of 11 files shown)


stringhandler (commented):
utACK 96074713b58c408e864e0144f0bfe27095f756e6

Looks good. Difficult to evaluate for safety, but review looks good.

nit: Any reason for naming it `exemplar` instead of `example`?

ivanlele:
> utACK [9607471](https://github.com/BlockstreamResearch/SimplicityHL/commit/96074713b58c408e864e0144f0bfe27095f756e6)
> 
> Looks good. Difficult to evaluate for safety, but review looks good.
> 
> nit: Any reason for naming it `exemplar` instead of `example`?

At first, I wanted to put this crate into examples, but decided against it because it mostly contained Symfony files. If it makes some sense, I can move it there.

I think I was thinking it was confusing, when you look at the root project structure, “external-jet-lib-exemple” would make sense to be in examples. So I’ve added “ar” to make it less eye-catching.

I’ll update to “example,” since I now also see that it makes a little sense.

ivanlele:
Rebased this one after #355 added `JetHinter::conjure`

KyrylR:
Could we add more guidance in rust-docs that explain the context?

apoelstra (approved):
ACK 832a441ae8d12a8ba8581e156cc015a666d1eda9; successfully ran local tests; cool!

KyrylR (approved):
ACK 832a441ae8d12a8ba8581e156cc015a666d1eda9; successfully ran local tests

Issue #327, opened 2026-05-21
Make `DependencyGraph` and `Module` public


gerau:
### Project

compiler

### Describe the feature

The main motivation behind this is to use them in the LSP without much code duplication. Because the LSP processes files using the internal tooling of the compiler's older version, this is currently not possible in version 0.6.0-rc.0, because `DependencyGraph` is private. Also, we need to somehow extract information about the linearization order and paths that are hidden behind each `file_id`, so it would be nice to make `Module` public as well (and make getter for `modules` field in `DependencyGraph`). 

I understand that exposing this interface might not be the best possible decision, so I would be happy to hear other solutions besides this.

LesterEvSe:
As I understand it, the main issue is that it is difficult to retrieve all the necessary information within the driver’s private functions.

This can be fixed by extending the `Span` field and adding `file_id: u32` to it, and replacing `lookup` and `paths` https://github.com/BlockstreamResearch/SimplicityHL/blob/f3168af1ff250ae179aa9db9002aeb3a26d65690/src/driver/mod.rs#L94-L102 in `driver.rs` with the following struct for example:

``` Rust
pub struct SourceMap {
    lookup: HashMap<CanonPath, usize>,
    paths: Vec<CanonPath>,
}
```

which can be made public and will be able to contain all the necessary files that can be accessed via `file_id`.

I think this is a more global issue related to changes in the `ErrorCollector` (one of the issue of this type: https://github.com/BlockstreamResearch/SimplicityHL/issues/290). Currently, it works well for detecting a single error in a single file, but it is poorly adapted to multiple errors across multiple files.

By addressing this issue, it can provide comprehensive error information without requiring the `DependencyGraph` or `Module` structs to be made public

PR #360, opened 2026-06-17
feat: SourceMap

gerau:
This adds `SourceMap`, which is actually a wrapper around `ids` and `paths` inside the driver (see https://github.com/BlockstreamResearch/SimplicityHL/issues/327#issuecomment-4508425248).
It also stores and exposes `SourceMap` and the built program in `TemplateProgram`. 

Motivation: It mainly exposes things for the LSP to access the parser program and the needed indexes to resolve `file_id` for functions.

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -56,6 +56,53 @@ struct SourceModule {
     program: parse::Program,
 }
 
+/// Record of all modules that was discovered.
+#[derive(Debug, Clone, Default)]
+pub struct SourceMap {
+    /// Fast lookup: `CanonPath` -> Module ID.
+    /// A reverse index mapping absolute file paths to their internal IDs.
+    /// This solves the duplication problem, ensuring each file is only parsed once.
+    ids: HashMap<CanonPath, usize>,
+
+    /// Fast lookup: Module ID -> `CanonPath`.
+    ///
+    /// A direct index mapping internal IDs back to their absolute file paths.
+    /// This serves as the exact inverse of the `lookup` map.
+    ///
+    /// This is highly useful for error reporting and diagnostics.
+    paths: Vec<CanonPath>,
+}
+
+impl SourceMap {
+    fn new(root_source: CanonSourceFile) -> Self {
+        let mut ids = HashMap::new();
+        ids.insert(root_source.name().clone(), MAIN_MODULE);
+        Self {
+            ids,
+            paths: vec![root_source.name().clone()],
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


gerau:
cc @LesterEvSe

src/driver/mod.rs
```diff
@@ -56,6 +56,53 @@ struct SourceModule {
...
+    pub fn get_path(&self, id: usize) -> Option<&CanonPath> {
+        self.paths.get(id)
+    }
+
+    pub fn insert(&mut self, path: CanonPath) -> usize {
+        let id = self.paths.len();
+        self.ids.insert(path.clone(), id);
+        self.paths.push(path);
+        id
+    }
```
LesterEvSe: I do not think we need to make this `pub`. Let's keep it private for now

src/lib.rs
```diff
@@ -56,6 +56,8 @@ pub struct TemplateProgram {
     simfony: ast::Program,
     file: Arc<str>,
     jet_hinter: Box<dyn ast::JetHinter>,
+    source_map: SourceMap,
+    driver_program: parse::Program,
```
LesterEvSe: Let's change it to `resolved_program`, because the name `driver_program` is not actually accurate. Propagate these changes throughout the following code

src/driver/mod.rs
gerau: fixed

src/lib.rs
gerau: changed

src/lib.rs
```diff
@@ -71,6 +73,7 @@ impl TemplateProgram {
     ) -> Result<String, String> {
         let mut error_handler = ErrorCollector::new();
         Self::dependency_helper(source, dependency_map, &mut error_handler)?
+            .0
             .ok_or_else(|| error_handler.to_string())
             .map(|p| p.to_string())
```
LesterEvSe:
```suggestion
        let (resolved_program, _) =
            Self::dependency_helper(source, dependency_map, &mut error_handler)?;

        resolved_program
            .ok_or_else(|| error_handler.to_string())
            .map(|p| p.to_string())
```

gerau: applied suggestion

LesterEvSe (approved):
ACK 3ef0b7560b63bd362f1e29a325eaf69968ccfea9; tested locally

KyrylR (approved):
ACK 3ef0b7560b63bd362f1e29a325eaf69968ccfea9

PR #361, opened 2026-06-22
lib: change error return type for TemplateProgram

gerau:
Last bit to make the LSP work with current version of the compiler. 

Motivation: in an earlier version of the LSP, the compilation process was fully copied from `TemplateProgram::new()`, so this would allow us to simply get all error information through TemplateProgram instead of building the program manually.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -418,6 +418,8 @@ impl Default for ErrorCollector {
     }
 }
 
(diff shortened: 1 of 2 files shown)


KyrylR (approved):
ACK 27ca31e3a99963cc63b04a2730f2db64e19e39c5; successfully ran local tests

PR #363, opened 2026-06-25
Move `file_id` into `Span`

### LINKED ISSUE #347: Refactor(ErrorHandler): Move `file_id` into `Span` and compute it during parsing
## Motivation

Currently, several AST structures in `parse.rs` (such as `TypeAlias` and `Function`) manually define a `file_id: usize` field to distinguish between identical structures across different files. It would be much cleaner to embed this `file_id` directly into the `Span` structure itself.

## The Problem

During the parsing stage, we currently initialize this `file_id` to a dummy value (0). Later, during the driver stage, we mutate the AST to inject the actual file ID generated by the driver. This approach feels like a crutch. It is highly inconvenient because it requires the driver to reach directly into the parsed structs to mutate them.

## The Solution

Instead of mutating the AST later, we should compute the `file_id` directly during the parsing stage. By hashing the specific file and assigning that hash as the `file_id` upfront, the driver will immediately have access to the correct ID. This significantly simplifies the driver's logic.

LesterEvSe:
This PR moves `file_id` into `Span` as a structural component of the source location. It also removes the unnecessary `file_id` and `set_file_id` methods from the `parse.rs` file.
Refactored the driver code a bit

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -85,11 +85,15 @@ impl SourceMap {
         }
     }
 
-    fn insert(&mut self, path: CanonPath) -> usize {
+    fn insert(&mut self, path: CanonPath) {
         let id = self.paths.len();
         self.ids.insert(path.clone(), id);
         self.paths.push(path);
-        id
+        debug_assert_eq!(self.ids.len(), self.paths.len());
+    }
+
+    fn len(&self) -> usize {
+        self.paths.len()
     }
 
     pub fn get_id(&self, path: &CanonPath) -> Option<usize> {
@@ -137,10 +141,10 @@ pub(crate) struct DependencyGraph {
     /// Fast bidirectional lookup between `CanonPath` and Module IDs.
     source_map: SourceMap,
 
-    /// Memoized results of [`crate::resolution::DependencyMap::resolve_path`] to avoid
-    /// resolving the same [`parse::UseDecl`] twice — once during the driver phase
-    /// and once during the building phase after linearization.
-    use_cache: HashMap<parse::UseDecl, ResolvedUse>,
+    /// Memoizes [`crate::resolution::DependencyMap::resolve_path_internal`] results,
+    /// keyed by the source [`Span`] of each `use` declaration, so the resolver runs
+    /// once per occurrence and the linearization phase can look the result up directly.
+    use_cache: HashMap<Span, ResolvedUse>,
 
     // TODO: Consider to optimising this with `Vec` instead of `HashMap`
     /// The Adjacency List: Defines the Directed acyclic Graph (DAG) of imports.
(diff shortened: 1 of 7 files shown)


src/driver/mod.rs
```diff
@@ -83,11 +83,15 @@ impl SourceMap {
         }
     }
 
-    fn insert(&mut self, path: CanonPath) -> usize {
+    fn insert(&mut self, path: CanonPath) {
         let id = self.paths.len();
         self.ids.insert(path.clone(), id);
```
KyrylR, line 90: Let's explicitly ignore returned value here

src/driver/mod.rs
```diff
@@ -83,11 +83,15 @@ impl SourceMap {
...
         self.paths.push(path);
-        id
+    }
+
+    /// Since `ids` and `paths` are equal, we can return len one of them
+    fn len(&self) -> usize {
+        self.paths.len()
```
KyrylR, line 96: Maybe we should have a guard here to make sure that `ids` and `paths` are equal?

src/driver/mod.rs
```diff
@@ -321,15 +293,21 @@ impl DependencyGraph {
...
+                import_span,
+                ctx.handler,
+            ) else {
                 // Safe to ignore output: previous `.contains` check prevents collisions.
                 let _ = ctx.invalid_imports.insert(path);
                 continue;
             };
 
-            let new_id = self.source_map.insert(path.clone());
+            self.source_map.insert(path.clone());
```
KyrylR, line 315: Let's explicitly ignore here too

src/driver/mod.rs
```diff
@@ -340,8 +318,41 @@ impl DependencyGraph {
...
+        let source = CanonSourceFile::new(path.clone(), Arc::from(content));
+        let ast =
+            parse::Program::parse_from_str_with_errors(new_id, source.clone(), &mut error_handler);
+
+        if error_handler.has_errors() {
+            handler.extend_with_handler(source, &error_handler);
+            None
+        } else {
+            ast.map(|program| SourceModule { source, program })
+        }
```
KyrylR, line 363:
```suggestion
        if error_handler.has_errors() {
            handler.extend_with_handler(source, &error_handler);
            return None;
        }
        
        ast.map(|program| SourceModule { source, program })        
```

src/driver/mod.rs
```diff
@@ -350,11 +361,31 @@ impl DependencyGraph {
...
 
 impl<'a> ImportContext<'a> {
+    /// Recursively walks an item, collecting resolved imports.
+    /// Recurses into inline `mod` blocks.
+    fn process_item(&mut self, item: &parse::Item, valid_imports: &mut Vec<(CanonPath, Span)>) {
+        match item {
+            parse::Item::Use(use_decl) => {
+                if let Some(import) = self.resolve_single(use_decl) {
+                    valid_imports.push(import);
+                }
```
```suggestion
                imports.extend(self.resolve_single(use_decl));
```

src/driver/mod.rs
LesterEvSe: Currently, this function has no return value

apoelstra:
37ab5429c20ddfa3a912a745542363dc07486484 needs rebase

src/driver/mod.rs
LesterEvSe: Redundant, as explained above

src/driver/mod.rs
LesterEvSe: Add `debug_assert_eq!` inside the `insert` function to verify this

apoelstra:
ebf0d21f3a97ee726fa9c97ae4beebc412636ff8 needs rebase

apoelstra:
In c81d8d0ae97f290f143f629c25bf37eb9bc0358a:

Can wait for another PR, but we should also change `Span::new` to take a `usize` for the file_id and a `std::ops::Range` rather than two `usizes` for the span. This should go a long way toward preventing mistakes when calling the constructor. It would also emphasize that the span's range is half-open.

Sadly we can't store a range inside `Span` because range isn't `Copy`, for dumb historic reasons. But we can at least take one in the constructor.

LesterEvSe:
Hm, interesting; imo it would be better to make `Iterator` `Copy`, but it is what it is. I will change the constructor for this

apoelstra (approved):
ACK 7b6db6cbcf56c340ac66635ae13026e56be2f510; successfully ran local tests

KyrylR (approved):
ACK 7b6db6cbcf56c340ac66635ae13026e56be2f510; successfully ran local tests

PR #364, opened 2026-06-25
Add a target option for simc

### LINKED ISSUE #224: Overriding the Elements jet set with the `--target` parameter

ivanlele:
This PR creates a new compilation option to choose a target jet set. Closes #224

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,8 +2,10 @@ use base64::display::Base64Display;
 use base64::engine::general_purpose::STANDARD;
 use clap::{Arg, ArgAction, Command};
 
-use simplicityhl::ast::ElementsJetHinter;
+use simplicityhl::ast::{CoreJetHinter, ElementsJetHinter, JetHinter};
 use simplicityhl::error::should_color;
+#[cfg(feature = "external-jets")]
+use simplicityhl::jet::external::{init_external_jet_lib, ExternalJetHinter};
 use simplicityhl::version::SimcDirective;
 use simplicityhl::{
     resolution::DependencyMapBuilder, source::CanonPath, source::CanonSourceFile, AbiMeta,
@@ -11,6 +13,7 @@ use simplicityhl::{
 };
 use simplicityhl::{UnstableFeature, UnstableFeatures};
 use std::path::Path;
+use std::str::FromStr;
 use std::{env, fmt, io};
 
 #[cfg_attr(feature = "serde", derive(serde::Serialize))]
@@ -28,6 +31,9 @@ struct Output {
     /// versions can produce different CMRs from the same source, so the version
     /// travels with the artifact as metadata (it is not part of the program).
     compiler_version: &'static str,
+    /// Compilation target used to select the jet set. Different targets can
+    /// produce different CMRs from the same source and compiler version.
+    target: String,
 }
 
 impl fmt::Display for Output {
@@ -35,6 +41,7 @@ impl fmt::Display for Output {
         writeln!(f, "Program:\n{}", self.program)?;
         writeln!(f, "CMR:\n{}", self.cmr)?;
         writeln!(f, "Compiler version:\n{}", self.compiler_version)?;
+        writeln!(f, "Target:\n{}", self.target)?;
         if let Some(witness) = &self.witness {
             writeln!(f, "Witness:\n{}", witness)?;
         }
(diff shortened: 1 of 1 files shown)


KyrylR:
needs rebase

apoelstra:
416a8dfcb01cb94b64b454c53f0228a998a91171 needs rebase

ivanlele:
Rebased

ivanlele:
Made some improvements per @KyrylR's suggestions

KyrylR (approved):
ACK 235a8144e50d03d1ef7fe4ac130ab9ec9cec61a4; successfully ran local tests, code review

KyrylR:
needs rebase

KyrylR:
cc @delta1 

@ivanlele needs rebase again

src/main.rs
```diff
@@ -45,6 +48,82 @@ impl fmt::Display for Output {
...
+impl Target {
+    #[allow(clippy::unnecessary_wraps)]
+    fn jet_hinter(&self) -> Result<Box<dyn JetHinter>, Box<dyn std::error::Error>> {
+        match self {
+            Target::Elements => Ok(Box::new(ElementsJetHinter::new())),
+            Target::Core => Ok(Box::new(CoreJetHinter::new())),
+            #[cfg(feature = "external-jets")]
+            Target::External(path) => {
+                unsafe {
+                    init_external_jet_lib(path)?;
```
apoelstra: In c0772e0d734f4d497aff7ca2e98aa1f83d7822f8:

This unsafe code needs a `SAFETY:` comment explaining why it's sound.

Also, this unsafe code is not sound. You need to do some sort of validation on `path` before you can call `init_external_jet_lib`.

ivanlele: I pushed changes to resolve this

apoelstra: `The user MUST trust the library to implement the external jet ABI.` is not a reasonable safety comment. We cannot ship code that exhibits UB based on user-provided input.

ivanlele: I suppose, I've written this in the wrong way. What I meant with this, that the dynamic library carries arbitrary binary code and we are not responsible for what it will decide to execute.

I do say "Initialization verifies that every required symbol exists" a line below and it actually does, so it should eliminate UB.

Maybe it would be better to just remove this:
"The user MUST trust the library to implement the external jet ABI."

As we do check that ABI is present.

apoelstra: In 7172f08c40c2d481182672fb68af4f746decbe29:

I think we should move `validate_external_jet_lib_path` to right before the unsafe call. As written, `Target` has a public `Target::External(string)` constructor, but the SAFETY comment claims that the string has been validated. We can't enforce that without changing the enum or moving the check.

I also wonder whether we need to use `String` or `str` here at all, since it imposes a UTF-8 requirement on paths that we shouldn't need to.

src/main.rs
```diff
@@ -45,6 +48,129 @@ impl fmt::Display for Output {
...
+impl FromStr for Target {
+    type Err = TargetParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "elements" => Ok(Target::Elements),
+            "core" => Ok(Target::Core),
+            #[cfg(feature = "external-jets")]
+            _ if s.starts_with("external:") => {
+                let path = s.trim_start_matches("external:");
```
stringhandler: in 667d0b3302e9ac820b54b9ed32599623a1784471

`strip_prefix` is more suitable than `trim_start_matches` here. trim_start_matches removes multiple occurrences

src/main.rs
```diff
@@ -45,6 +48,129 @@ impl fmt::Display for Output {
...
+    Elements,
+    Core,
+    #[cfg(feature = "external-jets")]
+    External(String),
+}
+
+#[derive(Debug)]
+struct TargetParseError(String);
+
+impl fmt::Display for TargetParseError {
```
stringhandler, line 71: LLM finding:

The Display impl is never reached — errors print via Debug 

main returns Result<(), Box<dyn Error>>, and Rust's Termination impl formats the error with Debug, not Display. Because TargetParseError uses #[derive(Debug)], the hand-written Display at line 62 is dead for every error that escapes jet_hinter(), and the user sees the derived tuple-struct form:


$ simc --target external:./foo.dylib examples/cat.simf
Error: TargetParseError("Invalid external target path './foo.dylib': No such file or directory (os error 2)")
Implementing Debug by hand to delegate to Display fixes it in one place:


impl fmt::Debug for TargetParseError {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        fmt::Display::fmt(self, f)
    }
}
Related: because path validation lives in jet_hinter() rather than from_str, the two failure modes for the same flag are formatted completely differently — external: (empty) gets clap's error: invalid value 'external:' for '--target <TARGET>': …, while external:./missing.dylib gets the raw Error: TargetParseError(…) above. Moving validate_external_jet_lib_path into from_str would make both clap-formatted and would fail fast before any source file is read. The dlopen itself would stay in jet_hinter().

src/main.rs
```diff
@@ -45,6 +48,129 @@ impl fmt::Display for Output {
...
+    })?;
+
+    if !canonical_path.is_file() {
+        return Err(TargetParseError(format!(
+            "Invalid external target path '{}': path is not a regular file",
+            canonical_path.display()
+        )));
+    }
+
+    let expected_extension = std::env::consts::DLL_EXTENSION;
```
stringhandler: LLM finding:

The extension check rejects loadable libraries 
std::env::consts::DLL_EXTENSION is a single value per platform (dylib on macOS, so on Linux, dll on Windows), but dlopen accepts more than that:

macOS loads .so files fine — this is exactly what you get from a library built on Linux, or from build systems that emit .so everywhere. Currently rejected:

$ simc --target external:/tmp/lib.so examples/cat.simf
Error: … expected a .dylib shared library
Versioned Linux sonames are rejected — Path::extension() on libfoo.so.1 returns "1", so the canonical libfoo.so.1 form fails on the platform where it's most common.
Since the real safety gate is Library::load (which fails loudly on a non-library) plus the symbol resolution in ExternalJetLib::load, this check buys little and costs valid inputs. I'd either accept the set {so, dylib, dll} regardless of host, or drop the check and let the loader produce the error.

stringhandler (commented):
I think the feature gating is ok, but I would say that we should always print the options in clap, but rather error if they are used without the correct compiled features. 


Some other LLM findings:

Output records compiler_version but not the target, and the target changes the CMR

The doc comment on compiler_version at line 30 gives the rationale for carrying it: "Different compiler versions can produce different CMRs from the same source, so the version travels with the artifact as metadata."

That reasoning now applies verbatim to --target. Same source, same compiler, different target:


$ simc examples/cat.simf         # CMR e65e19e139a13583a0a7efb24be13c20d578f06f51b2a7fe7c7b9097072dbabe
$ simc -t core examples/cat.simf # CMR c83aea0e102548c2a441c6d4d268fb0469263601ef705ea61fa18882938ec06f

With --json consumers now unable to tell which jet set an artifact was built against, a target: String field on Output seems like it belongs in this PR. (It would need Display/Debug on Target, which it doesn't currently derive.)

ivanlele:
I iterated on all comments that been written here

src/main.rs
stringhandler: @ivanlele Some of these comments still need to be addressed.

src/main.rs
stringhandler: resolved

src/main.rs
stringhandler: resolved

src/main.rs
stringhandler: resolved

src/main.rs
stringhandler: The most common, honest problem here will be if we change the interface and then old libraries mismatch. Perhaps we should make a call to get the external jet ABI version that the library was made for to check for mismatches. This doesn't stop malicious libraries, but will prevent common mistakes.

apoelstra: In fc4f030b60a258b51d567df3139e3e0e6add7aff:

The problem is still here. If this is the logic we want, then `jet_hinter` needs to be marked `unsafe` and we need a `SAFETY` comment at its call sites.

apoelstra:
1320c9a56c9a3f3644117857973bec4c3e0eb9ae needs rebase

ivanlele:
> [1320c9a](https://github.com/BlockstreamResearch/SimplicityHL/commit/1320c9a56c9a3f3644117857973bec4c3e0eb9ae) needs rebase

Done. I also tried to address your latest comment here

PR #367, opened 2026-07-01
Wasm external jets support

ivanlele:
This pull request adds support for external jet library loading via a WASM module provided by the host as a plugin for a WASM instance of the compiler.

This feature is useful when Simplicity source is being compiled against external jets in a browser environment (e.g. SimplicityHL Web IDE, Simplicity Unchained Web).


external-jet-lib-example/js-host/run.mjs
```diff
@@ -0,0 +1,178 @@
+import { readFile } from "node:fs/promises";
```
KyrylR, line 1: Is there a reason to use JS over TS?

KyrylR (approved):
The configuration part looks quite brittle and complex 

Over time (when/if there are more use cases for this functionality) hopefully it would improve 

ACK 7414d98cbaabf5687ccf0abd9270eaf7767026ab; code review, successfully ran local tests

external-jet-lib-example/js-host/run.mjs
ivanlele: None, actually. It's just what came to mind first.

KyrylR:
Useful for the external jets (for projects like Simplicity Unchained)

src/jet/external/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+            .expect("External jet lib is not initialized. Please call init_external_jet_lib first.")
+            as &dyn ExternalJetLib
+    }
+}
+
+/// External jet integration interface.
+///
+/// It is used by different loadable libraries backend to connect their
+/// implementations of required jets.
+pub trait ExternalJetLib {
```
stringhandler, line 88: I would prefer this file to be split up. It's quite messy with the WASM and other feature stuff. It will also be easier to find this trait in it's own file

src/jet/external/mod.rs
```diff
@@ -0,0 +1,336 @@
+#[cfg(not(target_arch = "wasm32"))]
+mod dynamic;
```
stringhandler, line 2: I wonder if the dynamic ability should be feature gated?

src/jet/external/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+pub fn deserialize_source_jet_classification(bytes: &[u8]) -> Option<SourceJetClassification> {
+    let (&tag, mut rest) = bytes.split_first()?;
+    match tag {
+        0 => Some(SourceJetClassification::Unary),
+        1 => Some(SourceJetClassification::Binary),
+        2 => Some(SourceJetClassification::Ternary),
+        3 => Some(SourceJetClassification::Quaternary),
+        4 => {
+            let count = read_u32(&mut rest)? as usize;
+            let mut types = Vec::with_capacity(count);
```
stringhandler, line 269: LLM Finding:

count is read straight from the plugin's buffer, so a malformed (or hostile) message can request a ~4 GiB allocation here before the loop below bails on the first truncated read. Worth bounding it — on the native path the plugin is arbitrary in-process code anyway, but on wasm it's genuinely sandboxed, and this is one of the places that decides whether a bad plugin can abort the compiler.

Minimum encoding per element is 4 bytes (the u32 length), so:


let mut types = Vec::with_capacity(count.min(rest.len() / 4));

src/jet/external/mod.rs
stringhandler: LLM finding:

This trait is public, but neither implementor is reachable from outside the crate — mod dynamic; (L2) and mod wasm; (L5) are both private, and init_external_jet_lib only ever installs the built-in backend for the current target. So a downstream user can name ExternalJetLib and implement it, but has no way to register it, which makes the doc comment's "used by different loadable libraries backend" true only within the crate.

Either pub(crate) it, or add a registration entry point that accepts a user-supplied backend. A public trait with no public use is harder to remove later than to not export now.

src/jet/external/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+/// the library is loaded from the specified path.
+///
+/// # Safety
+///
+/// The caller must ensure that the loaded library exports each of the
+/// symbols listed below with signatures matching the corresponding
+/// fields of [`ExternalJetLib`]. Calling a function through a
+/// mismatched signature is undefined behavior.
+#[cfg(target_arch = "wasm32")]
+pub unsafe fn init_external_jet_lib() -> Result<(), Box<dyn std::error::Error>> {
```
stringhandler, line 38: LLM finding:

ExternalJetWasmLib::load() (wasm.rs:30) returns a unit struct and cannot fail, so this OnceLock isn't protecting any state — but it does create a state that can be wrong: skip init and external_jet_lib() panics with "not initialized" even though on wasm there is nothing to initialize (the host wired the imports before any Rust ran).

The cost shows up in the example: compiler_wasm.rs:41 has to do if msg.contains("already been initialized") { Ok(()) }, i.e. string-match an error message to recover from a non-error.

Suggest dropping the OnceLock on the wasm path — external_jet_lib() can return &ExternalJetWasmLib from a const — and either making init_external_jet_lib() a no-op for source compatibility or omitting it on wasm entirely. Worth settling now since it's public API.

src/jet/external/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+use crate::jet::{JetHL, SourceJetClassification, TargetJetClassification};
+use crate::parse::ParseFromStr;
+use crate::types::AliasedType;
+
+#[cfg(target_arch = "wasm32")]
+use crate::jet::external::wasm::ExternalJetWasmLib;
+#[cfg(not(target_arch = "wasm32"))]
+use crate::jet::external::{dynamic::ExternalJetDynamicLib, loaders::dynlib::Library};
+
+/// Load the external jet library. When compiled for wasm32, the library is loaded
```
stringhandler, line 27: LLM finding:

This doc block looks copied from the native version — it says the caller must ensure the loaded library exports matching symbols, but on wasm the caller can't. There is no library and no path; the JS host supplied the imports at instantiation, and by the time this runs a signature mismatch has already either failed instantiation or silently succeeded.

The real obligation belongs to the embedder: the host must supply imports under the module name simplicityhl-plugin conforming to the shim ABI declared in wasm.rs:36-48. Worth restating in those terms, since this is the # Safety section people will actually rely on.

src/jet/external/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+        Box::new(ExternalJetHinter)
+    }
+
+    fn conjure(&self, jet: &dyn Jet) -> Option<Box<dyn JetHL>> {
+        let container = external_jet_lib();
+        container.conjure(jet)
+    }
+}
+
+/// Serialize a [`SourceJetClassification`] into a portable byte buffer.
```
stringhandler, line 231: LLM finding:

Nothing in the repo exercises this format in both directions. I checked it by hand against bool, List<bool, 4>, [bool; 3], Either<bool, bool>, (bool, bool) and Pubkey and all six round-trip — but that's the sort of property that should be pinned, especially as it rests on AliasedType's Display staying in sync with the parser, which nothing currently enforces.

A unit test over a handful of types would do; better still, the crate already derives Arbitrary for AliasedType under the arbitrary feature, so a proptest asserting deserialize(serialize(x)) == Some(x) is only a few lines.

src/jet/external/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+        1 => Some(TargetJetClassification::Custom(read_aliased_type(
+            &mut rest,
+        )?)),
+        _ => None,
+    }
+}
+
+fn write_aliased_type(out: &mut Vec<u8>, ty: &AliasedType) {
+    let s = ty.to_string();
+    out.extend_from_slice(&(s.len() as u32).to_le_bytes());
```
stringhandler, line 314: LLM finding:

s.len() as u32 here and types.len() as u32 at L247 are unchecked. Unreachable in practice — it'd need a 4 GiB type name — but the root Cargo.toml sets the convention that "every cast should contain a code comment" alongside the cast_possible_truncation = "allow". A one-line comment on each would satisfy it.

src/jet/external/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+
+    fn conjure(&self, jet: &dyn Jet) -> Option<Box<dyn JetHL>> {
+        let container = external_jet_lib();
+        container.conjure(jet)
+    }
+}
+
+/// Serialize a [`SourceJetClassification`] into a portable byte buffer.
+///
+/// Layout:
```
stringhandler, line 233: LLM finding/question:

Design question rather than a defect: the tag byte is doing double duty as an implicit version. If SourceJetClassification gains a variant later, an old plugin paired with a new compiler disagrees silently rather than failing cleanly.

Not a regression — the native path has no versioning at all — but the wasm path is the one where an independently built and independently shipped plugin is a realistic scenario, so it's where a version byte or handshake starts to earn its keep. Is that in scope here, or deliberately deferred?

src/jet/external/mod.rs
stringhandler: Seems like a nit

src/jet/external/wasm.rs
```diff
@@ -0,0 +1,193 @@
...
+extern "C" {
+    fn cmr(index: u64, out_ptr: *mut u8, cap: usize) -> i32;
+    fn source_ty(index: u64, out_ptr: *mut u8, cap: usize) -> i32;
+    fn target_ty(index: u64, out_ptr: *mut u8, cap: usize) -> i32;
+    fn encode(index: u64, out_ptr: *mut u8, cap: usize) -> i32;
+    fn display(index: u64, out_ptr: *mut u8, cap: usize) -> i32;
+    fn source_jet_classification(index: u64, out_ptr: *mut u8, cap: usize) -> i32;
+    fn target_jet_classification(index: u64, out_ptr: *mut u8, cap: usize) -> i32;
+
+    fn cost(jet: ExternalJet) -> Cost;
```
stringhandler, line 46: LLM finding (blocking):

These three are declared extern "C" here, but the plugin defines them with the Rust ABI — external-jet-lib-example/src/lib.rs:188, :297, :310 are plain #[no_mangle] pub fn with no #[cfg(target_arch = "wasm32")] variant and no extern "C". Every other entry point in this PR got a proper extern "C" shim; these were left as the untouched native definitions and are being imported under a different ABI. The #[allow(improper_ctypes)] on L36 is what suppresses the diagnostic.

It happens to work right now: Cost is pub struct Cost(u32) and ExternalJet wraps a single u64, both single-field newtypes without #[repr(C)], and on wasm32-unknown-unknown the Rust and C ABIs both lower those to a bare scalar. But the Rust ABI is explicitly unspecified, Cost belongs to an upstream crate that could gain a field, and the entire premise of the wasm path is that these two modules are compiled independently — so "same toolchain, same lowering" is exactly the assumption we shouldn't be leaning on here.

Suggest giving all three extern "C" shims on the plugin side like the rest, and narrowing the improper_ctypes allow to whatever genuinely still needs it

stringhandler (commented):
Sorry for so many comments, I went through a detailed process with an LLM explaining the bits of the PR for me because I've fallen behind on the external jets loading and asked it to review at the same time.

Most of the LLM findings are optional, but it did highlight a potential future breaking problem which I've marked "(blocking)".

Not included in this PR, but the CI job for WASM needs to be changed from

```
- run: cargo check --target wasm32-unknown-unknown
```
to include the `external jets` feature.

```
- run: cargo check --target wasm32-unknown-unknown --features external-jets
```

ivanlele:
> Sorry for so many comments, I went through a detailed process with an LLM explaining the bits of the PR for me because I've fallen behind on the external jets loading and asked it to review at the same time.
> 
> Most of the LLM findings are optional, but it did highlight a potential future breaking problem which I've marked "(blocking)".
> 
> Not included in this PR, but the CI job for WASM needs to be changed from
> 
> ```
> - run: cargo check --target wasm32-unknown-unknown
> ```
> 
> to include the `external jets` feature.
> 
> ```
> - run: cargo check --target wasm32-unknown-unknown --features external-jets
> ```

Thanks for the review! It will take me some time to go over this, but I already see points from your comments that really need to be addresed

ivanlele:
@stringhandler, I've addressed your comments

PR #368, opened 2026-07-03
Store file content inside the `SourceMap` struct

LesterEvSe:
Follow the same approach as Rust's `SourceMap` ([source code](https://github.com/rust-lang/rust/blob/c397dae808f70caebab1fc4e11b3edf7e59f58c7/compiler/rustc_span/src/source_map.rs#L1-L10)): the map should own the source bytes of every file it tracks, not just their paths.

Storing content indexed by `file_id` lets a future Diagnostic retrieve the specific lines where errors occurred using `ariadne` render

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -72,36 +72,48 @@ pub struct SourceMap {
     /// This serves as the exact inverse of the `lookup` map.
     ///
     /// This is highly useful for error reporting and diagnostics.
-    paths: Vec<CanonPath>,
+    entries: Vec<CanonSourceFile>,
 }
 
 impl SourceMap {
     fn new(root_source: CanonSourceFile) -> Self {
         let mut ids = HashMap::new();
+
         ids.insert(root_source.name().clone(), MAIN_MODULE);
+
         Self {
             ids,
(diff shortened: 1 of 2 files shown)


src/driver/mod.rs
```diff
@@ -72,36 +72,48 @@ pub struct SourceMap {
...
+        debug_assert_eq!(self.ids.len(), self.entries.len());
     }
 
     fn len(&self) -> usize {
-        self.paths.len()
+        self.entries.len()
     }
 
-    pub fn get_id(&self, path: &CanonPath) -> Option<usize> {
+    pub fn id(&self, path: &CanonPath) -> Option<usize> {
```
KyrylR, line 103: Is it Ok to break API here?

KyrylR (commented):
ACK 8a806c016aa725d9ac161554b7a020deb5154813 with comments; successfully ran local tests, code review

src/driver/mod.rs
LesterEvSe: I think @gerau can answer more precisely about this

gerau: yeah, it's fine

KyrylR (approved):

PR #376, opened 2026-07-17
Enums

KyrylR:
This PR is re implementation of https://github.com/BlockstreamResearch/SimplicityHL/pull/336 with the emphasis on nominative enums


KyrylR:
I do not have a complete understanding of Simplicity's type algebra, so I did a few iterations with Fable 5 to understand how it works

Here is the summary:

Simplicity's type algebra has exactly three constructors: the unit type `1`, products `A × B`, and **binary** sums `A + B`. There is no n-ary sum. A SimplicityHL enum with `n` variants means "exactly one of these `n` payloads" — so before anything can be compiled, the compiler must express that as some nesting of two-way `Either`s. `balanced_sum` is where that nesting is *chosen*: it folds the `n` variant payload types into one canonical tree of `Either`s.

The choice being **canonical** matters more than what the choice is. Four independent pieces of the compiler navigate this tree, and they only interoperate because all four recurse with the same split:

- `balanced_sum` (types.rs) — builds the structural *type*;
- `inject_variant` (compile) — construction wraps a payload in `injl`/`injr` down the path to its leaf;
- `case_tree` + `EnumMatch::compile` — a match becomes nested `case` combinators of the same shape;
- `enum_injection` / `as_enum_leaf` (value.rs) — witness values are encoded to, and decoded from, the same paths.

If any one of them split differently, a constructed `Action::Cold` would land on a leaf that the match's `case` tree interprets as a different variant. And because the leaf-position of a variant *is* its wire encoding, the shape is also part of the ABI — which is why the ABI ships variants strictly in declaration order.

## Theoretical foundation

**Finite coproducts from binary ones.** In type-theoretic terms an enum is a sum type; categorically, a coproduct `A₁ + A₂ + … + Aₙ`. A category with binary coproducts automatically has all finite *nonempty* coproducts, by iteration — and associativity holds up to canonical isomorphism: `(A+B)+C ≅ A+(B+C)`. So *semantically*, any bracketing represents the enum equally well; all bracketings are isomorphic. The compiler's job is merely to fix one associate and use it everywhere. `balanced_sum` is exactly the constructive content of that "finite coproducts exist" theorem: a fold of the binary coproduct over the leaf list.

**Why nonempty (the panic).** The empty coproduct is the initial object `0` — the uninhabited type. Simplicity's algebra is `{1, ×, +}` with no `0`: an uninhabited type is inexpressible (every Simplicity type is inhabited). Hence the `expect("at least one leaf")` here, and the matching non-empty invariant on `EnumInfo::new`. A one-leaf "sum" degenerates to the payload itself — a single-variant enum is just a nominal wrapper.

**Injections and case analysis.** A coproduct comes with injections `injᵢ : Aᵢ → ΣA` and the universal property that any family of maps `fᵢ : Aᵢ → C` factors through a unique copairing `[f₁,…,fₙ]`. Simplicity gives you exactly the binary versions: `injl`/`injr` are the two injections, and `case f g` is the binary copairing. Under the balanced encoding, the n-ary injection `injᵢ` becomes the composite of `injl`/`injr` steps along the path to leaf `i` (that's `inject_variant`), and the n-ary copairing becomes the tree of `case` nodes folded over the match arms (that's `case_tree`). The enum feature adds *no* new primitive — it's pure derived structure over Simplicity's existing algebra, which is also why an undeclared variant is unrepresentable: there is no fifth path through a four-leaf tree.

**The discriminant is the path.** A Simplicity sum value is encoded as a tag bit choosing left/right, then the payload. So a variant's encoding starts with the sequence of tag bits spelling its root-to-leaf path — the path *is* the discriminant; no separate tag field exists. This is where "why balanced, not a right-nested comb `A + (B + (C + …))`" gets its answer:

- **Comb:** variant `k` sits at depth `k+1` (except the last), so its tag costs ~`k` bits and matching it traverses `k` nested `case` nodes — a *unary* discriminant. A 256-variant enum's last variant pays 255 tag bits and 255 case nodes.
- **Balanced:** every leaf sits at depth `⌊log₂ n⌋` or `⌈log₂ n⌉`, so every variant's tag is ~`log₂ n` bits and every match dispatch traverses ~`log₂ n` cases — effectively a *fixed-width binary* discriminant, the same thing a hand-rolled tag field would give you, but derived instead of stored. (Information-theoretically: with no reason to favor any variant, equal-weight leaves make the complete binary tree the optimal prefix code — Huffman with uniform weights.)

**The exact shape.** `BTreeSlice` splits `n` leaves at `btree_split_index(n) = n − 2^(⌈log₂ n⌉−1)`: the right subtree always gets the largest power of two ≤ `n−1` leaves… concretely, the right child is a *perfect* tree and the left child absorbs the remainder, recursively. For `n = 3` that's `1 + (1+1)` — matching the `enum_structural_type_is_balanced_sum` test's expected `Either<1, Either<1,1>>`. Crucially, this is the *same* fold the codebase already uses for products (tuples and arrays are balanced products of the same `BTreeSlice` shape), so sums and products are duals built by one shape rule — one invariant to keep in sync instead of two.

One last practical point hidden in the doc comment: when a value is *constructed*, only the leaf actually inhabited carries data — the types of all sibling subtrees along the path aren't determined by the injection itself. They get pinned by Simplicity's type inference when the value meets its consumer (the match's `case` tree or the declared witness type), which is why `inject_variant` can build the path without ever spelling out the full enum type.

src/unstable.rs
```diff
@@ -265,6 +273,14 @@ impl FromStr for UnstableFeature {
...
 ///     }
 /// }
+///
+/// // Struct that gates a feature but has no children to recurse into.
+/// // The `recurse` clause is optional and may be omitted entirely.
+/// impl_require_feature! {
+///     EnumDeclaration {
+///         requires: UnstableFeature::Enums, span: span;
+///     }
+/// }
```
LesterEvSe, line 283: Do we need this?

KyrylR: Yes, to show that recurse is now optional

KyrylR:
Notes:

1. compilation with flattening turned on is broken (will be fixed by @LesterEvSe)
2. casting currently does not work between enums even if those casts are "safe" from types PoV (will be fixed in follow up PRs)

LesterEvSe:
`AbiMeta` serializes every type as its display string (`impl Serialize for ResolvedType`, `src/serde.rs`). For non-enum types that string fully determines the bit layout, so the JSON ABI is self-contained. For enums it degrades to the bare declared name (`"ACTION": "Action"`) with no variants, no payload types, no order.
This means two different programs can produce an identical ABI JSON entry `"ACTION": "Action"`, but because a variant is encoded by its position in the declaration:

``` Rust
// v1
enum Action { Inherit(u32), ColdSpend(u32) }
// v2
enum Action { ColdSpend(u32), Inherit(u32) }
```

they end up with a different CMR and different witness bytes. For example, the same witness value `{ "ACTION": "Action::Inherit(1)" }` encodes as different witnesses under v1 and v2. The ABI diff shows no change, while the wire format and the program identity have both changed.

So I think this needs to be described in the `architecture.md` file, or an issue should be opened to add more context to the `.wit` file for the user.

LesterEvSe:
Also, let's open an issue for fuzzing tests: currently fuzzing covers enum declarations only; construction, match/payload binding, and serde remain unfuzzed.

KyrylR:
> `AbiMeta` serializes every type as its display string (`impl Serialize for ResolvedType`, `src/serde.rs`). For non-enum types that string fully determines the bit layout, so the JSON ABI is self-contained. For enums it degrades to the bare declared name (`"ACTION": "Action"`) with no variants, no payload types, no order. This means two different programs can produce an identical ABI JSON entry `"ACTION": "Action"`, but because a variant is encoded by its position in the declaration:
> 
> ```rust
> // v1
> enum Action { Inherit(u32), ColdSpend(u32) }
> // v2
> enum Action { ColdSpend(u32), Inherit(u32) }
> ```
> 
> they end up with a different CMR and different witness bytes. For example, the same witness value `{ "ACTION": "Action::Inherit(1)" }` encodes as different witnesses under v1 and v2. The ABI diff shows no change, while the wire format and the program identity have both changed.
> 
> So I think this needs to be described in the `architecture.md` file, or an issue should be opened to add more context to the `.wit` file for the user.

We had the same behavior when updating the logic of the program (if you add more logic to the program, CMR will change but .wit file won't)

KyrylR:
> Also, let's open an issue for fuzzing tests: currently fuzzing covers enum declarations only; construction, match/payload binding, and serde remain unfuzzed.

Yep

LesterEvSe:
> We had the same behavior when updating the logic of the program (if you add more logic to the program, CMR will change but .wit file won't)

That make sense

LesterEvSe (approved):
ACK c242d0890cd2d07d9b7ace1725cb17693ee9db69; tested locally with `just check` and `just check_fuzz`

LesterEvSe:
@apoelstra, we merged it as is. The main problem is that enums currently do not work with dependencies and flattening. This will be fixed in follow-up PRs

apoelstra:
Went through c242d0890cd2d07d9b7ace1725cb17693ee9db69. This all looks good to me. There is a bit of followup:

* I suspect we'll want to significantly refactor our resolver/module/flattening system to handle enum variants as a form of exported symbol
* We ought to have the uninhabited type in SimplicityHL, and we can take empty enums to be uninhabited. It is impossible to write code that constructs a value of such a type, and any code that takes one as input is uncallable and we can skip compiling it. So it doesn't matter that Simplicity itself doesn't have an uninhabited type. But probably we need to overhaul our type system first

PR #378, opened 2026-07-22
Add multi-error support to `is_consistent` and print colored errors in `main.rs`

LesterEvSe:
- Output errors to `stderr` with color support in `main.rs`.
- Add multi-error support to `Arguments` and `Witness`.


KyrylR (approved):
ACK 38a4531a6ef8abe61775829d7800f53d67f76ab0; successfully ran local tests

PR #381, opened 2026-07-22
Add sorting to `is_consistent` function for deterministic error ordering

LesterEvSe:
Error output ordering was nondeterministic because `Arguments` and `Witness` used `HashMap` internally, causing diagnostics to be emitted in arbitrary iteration order across runs.

This PR add sorting to `is_consistent` for deterministic error ordering.


KyrylR (approved):
ACK 4a21dcd2940bde71850e2dde8585c46127a4a68c; 

nit: upd description pls as well

PR #383, opened 2026-07-28
Implement AST multi-error recovery

### LINKED ISSUE #205: Implement error states in parser
In #185 we used default values when we encountered error (for example, if we have an error in single expression, it return an unit type `()` ). 
It shouldn't return default values, but recovering to special error state, so an analyzer can distinguish between them, and handle it correctly. Right now if we have we don't return "poisoned" tree, but if we implement this we could analyze even broken code.

### LINKED ISSUE #207: Error recovery in analysis
Analysis should not stop at the first error, so we should implement error recovery and error collection. We can use a similar structure to #185, which would take a program and a collector and return an `Option`.

### LINKED ISSUE #348: Refactor(ErrorHandler): Enhance error diagnostics for use item resolution
We have a limitation where use statement diagnostics offer very basic error information without indicating which specific Item was expected to be loaded. To provide a broader error context, we should update the `resolve_processing_use_items_error` function in `ast.rs`. This tracking issue is related to the PR #337.

LesterEvSe:
Analysis now reports every independent error in one pass instead of stopping at
the first.

- **Poison, not fake values**; `Error` variants in the parse tree, AST, and type
  system (`TypeInner::Error`); recovery emits these instead of fabricated units,
  aliases, or sizes.
- **No cascades**; poisoned types are absorbed during analysis (`is_error` guards
  and `ResolvedType::compatible`), so one broken construct does not spawn secondary
  errors.
- **Collect, do not abort**; `Scope` carries a diagnostics sink; containers record
  an error and keep analyzing siblings, and `Program::analyze` returns `Option`.
- **Parser split**; `parse_from_str` (fragments, first error) vs
  `parse_from_content` (files, collect all), over one shared core.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -459,16 +459,56 @@ pub trait TypeDeconstructible: Sized {
 pub struct ResolvedType(TypeInner<Arc<Self>>);
 
 impl ResolvedType {
+    /// Diagnostic-facing compatibility: a poisoned type unifies with anything,
+    /// recursively. Use before emitting a type-mismatch error.
+    pub fn compatible(&self, other: &Self) -> bool {
+        use TypeInner::*;
+
+        match (self.as_inner(), other.as_inner()) {
+            (Never, _) | (_, Never) => true,
+            (Either(a, b), Either(c, d)) => a.compatible(c) && b.compatible(d),
+            (Option(a), Option(b)) => a.compatible(b),
+            (Tuple(a), Tuple(b)) => {
+                a.len() == b.len() && a.iter().zip(b.iter()).all(|(x, y)| x.compatible(y))
+            }
+            (Array(a, m), Array(b, n)) => m == n && a.compatible(b),
+            (List(a, m), List(b, n)) => m == n && a.compatible(b),
+            _ => self == other,
+        }
+    }
+
+    /// Replace this type's poisoned leaves with `other`'s types.
+    ///
+    /// Only meaningful for [`Self::compatible`] types, whose shapes line up. The
+    /// first concrete type to reach a poisoned slot fixes it, so a later
+    /// conflicting use of that slot is still a conflict rather than being
+    /// absorbed by the poison forever.
+    pub fn refine(&self, other: &Self) -> Self {
+        use TypeInner::*;
+
+        match (self.as_inner(), other.as_inner()) {
+            (Never, _) => other.clone(),
+            (Either(a, b), Either(c, d)) => Self::either(a.refine(c), b.refine(d)),
+            (Option(a), Option(b)) => Self::option(a.refine(b)),
+            (Tuple(a), Tuple(b)) if a.len() == b.len() => {
+                Self::tuple(a.iter().zip(b.iter()).map(|(x, y)| x.refine(y)))
+            }
+            (Array(a, m), Array(b, n)) if m == n => Self::array(a.refine(b), *m),
+            (List(a, m), List(b, n)) if m == n => Self::list(a.refine(b), *m),
+            _ => self.clone(),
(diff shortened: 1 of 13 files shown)


apoelstra:
In 763b2d81970d142e11b0a8a6f2035e3eae317764:

I think we should call the `Error` type "Uninhabited" (or "Never" if we want to be Rust-like), and similarly the `is_error` method should be `is_uninhabited` (or `is_never` or `is_empty` ... though I think `is_empty` would be confusing because it's such a common name in Rust).

The empty type has the properties you want -- it cannot be constructed, no code should ever be compiled which accepts it, and there is a unique function mapping it to every other type, and I'd like to add it to the language anyway, so we might as well add it here.

In 8a1b03c22f2b729d3180bf7c4d9fa128ee28834b:

Adds a blank line to witness.rs, otherwise doesn't touch the file.

LesterEvSe:
> In [763b2d8](https://github.com/BlockstreamResearch/SimplicityHL/commit/763b2d81970d142e11b0a8a6f2035e3eae317764):
> 
> I think we should call the `Error` type "Uninhabited" (or "Never" if we want to be Rust-like), and similarly the `is_error` method should be `is_uninhabited` (or `is_never` or `is_empty` ... though I think `is_empty` would be confusing because it's such a common name in Rust).

Changed `Error` to `Never` and `is_error` to `is_never` in `types.rs`

> In [8a1b03c](https://github.com/BlockstreamResearch/SimplicityHL/commit/8a1b03c22f2b729d3180bf7c4d9fa128ee28834b):
> 
> Adds a blank line to witness.rs, otherwise doesn't touch the file.

I accidentally refactored the code while reviewing it. Deleted it.

src/types.rs
```diff
@@ -31,6 +31,9 @@ pub enum TypeInner<A> {
     /// Nominal enum type, represented as a balanced sum of its variants'
     /// payload types
     Enum(EnumInfo),
+    /// Type of a recovered subtree. Compatible with every type; never re-reported.
```
apoelstra, line 34: In fdfae8b186ed724de436d0dc6b1428cdba340c54:

What is a "recovered subtree" what does it mean to "re-report" a type?

src/value.rs
```diff
@@ -937,6 +938,7 @@ impl crate::ArbitraryOfType for Value {
                     .collect::<arbitrary::Result<Vec<Self>>>()?;
                 Ok(Self::list(elements, ty.as_ref().clone(), *bound))
             }
+            TypeInner::Never => unreachable!("cannot generate a value of a poisoned type"),
```
apoelstra, line 941: In fdfae8b186ed724de436d0dc6b1428cdba340c54:

This one is not unreachable. We should return an Arbitrary error here indicating that there is no value of the never type.

LesterEvSe: Agreed with @KyrylR that this PR is too large to review as one unit. I'm splitting it into a sequence:

(Rough split, not yet validated)
1. Add the uninhabited `Never` type to `TypeInner` (+ `is_never`,
   `ResolvedType::compatible`, and the `Arbitrary` impl returning an error
   instead of `unreachable!`). Standalone and useful on its own.
2. `Error` variants in the parse tree + parser recovery (#205).
3. Diagnostics sink in `Scope`; containers collect instead of aborting (#207).
4. Split `parse_from_str` / `parse_from_content` by error policy.

@apoelstra, I plan to address your comment about the Arbitrary branch in (1). I'll also rewrite the doc comment "recovered subtree" and "re-report" are not defined anywhere. Sorry for the churn.

Keeping this open as a draft for the overall design; I'll close it once (4)
is merged.

PR #388, opened 2026-08-13
refactor: introduce `TemplateProgram` between AST and `CompiledProgram`

apoelstra:
This PR does a number of refactors ahead of introducing quoting and canonical CMRs. The goal of these changes is to rearrange compilation.

Rather than the process being

Templated AST (`TemplateProgram`) -> instantiated AST -> compiled program

instead it will be

Templated AST -> Templated compiled program -> compiled program

I tried to reduce churn but there are a couple renames that come with this:

1. `TemplateProgram` becomes `TemplateAst` (and a new `TemplateProgram` type is introduced)
2. `WitnessNames`, which is the name of the witness type for our compiled programs, becomes `TemplateProgramWitness`, and is expanded to track both witness names and parameters

In fact, `WitnessNames` is already (ab)used to store both witness names and parameters, which makes these refactors a little easier. But this appears to be a form a code reuse rather than a deliberate API choice; the current `WitnessName` struct is just a wrapper around `Arc<str>` that does not track what kind of object it's holding, and we distinguish based on what context it appears in. Because we instatiate before compiling, any remaining `WitnessNames` in a compiled program are definitely witness names and not parameter names. This PR changes that, so we need to extend the type.

Once these refactors are done, the actual rearranging of the instantiation step is suprisingly easy. But it will have to wait for the next PR.

This PR has no behavior changes, does not introduce any new functionality, and does not use any features of the upcoming rust-simplicity.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,7 +46,7 @@ use crate::error::DiagnosticManager;
 use crate::parse::ParseFromStrWithErrors;
 use crate::resolution::DependencyMap;
 use crate::source::CanonSourceFile;
-pub use crate::template_program::TemplateProgramWitness;
+pub use crate::template_program::{TemplateProgram, TemplateProgramWitness};
 pub use crate::types::ResolvedType;
 pub use crate::unstable::{UnstableFeature, UnstableFeatures};
 pub use crate::value::Value;
@@ -215,14 +215,14 @@ impl TemplateAst {
         }
 
         let commit = self.simfony.compile(
-            arguments,
+            arguments.shallow_clone(),
             include_debug_symbols,
             self.jet_hinter.clone_box(),
         )?;
 
         Ok(CompiledProgram {
             debug_symbols: self.simfony.debug_symbols(self.file.as_ref()),
-            simplicity: commit,
+            simplicity: commit.instantiate(arguments),
             witness_types: self.simfony.witness_types().shallow_clone(),
             parameter_types: self.simfony.parameters().shallow_clone(),
         })
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,7 @@ pub mod parse;
 pub mod pattern;
 pub mod resolution;
 pub mod source;
+pub mod template_program;
 pub mod unstable;
 
 #[cfg(feature = "serde")]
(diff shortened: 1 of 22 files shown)


stringhandler (approved):
ACK 9942a5a7c8ea0b8056d5c98a944f93fde5bfea90 Ran cargo test locally

src/template_program/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    /// Instantiates the templated program with the given arguments.
+    ///
+    /// ## Precondition
+    ///
+    /// The supplied `arguments` are consistent with the program's parameters.
+    /// Call [`Arguments::is_consistent`] before calling this method!
+    pub fn instantiate(self, _: Arguments) -> Arc<named::CommitNode> {
```
stringhandler, line 36: Strange to have the unused variable here. Perhaps we need a comment explaining its reason?

stringhandler (approved):
utACK 917c57213c709c0d1ce16a0f54a23be49fd9eea5

src/template_program/mod.rs
apoelstra: It will be used in the next PR.

stringhandler: Perfect

KyrylR (approved):
ACK 917c57213c709c0d1ce16a0f54a23be49fd9eea5

nit: `cargo test --locked --manifest-path ./bitcoind-tests/Cargo.toml` fails 

Not sure how to deal with this crate rn, though it would be nice if it were better integrated into the tests (if it is worth it)

src/lib.rs
```diff
@@ -45,26 +46,29 @@ use crate::error::DiagnosticManager;
 use crate::parse::ParseFromStrWithErrors;
 use crate::resolution::DependencyMap;
 use crate::source::CanonSourceFile;
+pub use crate::template_program::{TemplateProgram, TemplateProgramWitness};
 pub use crate::types::ResolvedType;
 pub use crate::unstable::{UnstableFeature, UnstableFeatures};
 pub use crate::value::Value;
+pub use crate::witness::{
+    Arguments, Parameters, WitnessNameToValueMap as _, WitnessTypes, WitnessValues,
+};
```
KyrylR: nit: it looks like now we have two exports of WitnessNameToValueMap, one below that is gated by serde feature, and this one

apoelstra: Oo, good catch. I'm not sure why I added this new export. It kinda looks like something my IDE would've done, but I don't think it'd ever add a `pub use` automatically.

apoelstra: Oh, I think actually it was just a rebasing mistake. Anyway, fixed.

stringhandler (approved):
utACK 0b58f68fd29856203566d317726175def34f8122

Issue #386, opened 2026-08-11
Is there no LICENSE in this repo?


apoelstra:
It appears there is no LICENSE or SPDX headers in this repo. We should add one. I propose we dual-license under MIT and Apache. See https://github.com/rust-bitcoin/rust-bitcoin/issues/5849 for some discussion about this.

cc @sanket1729 @uncomputable since this has significant code contributions from both of you.

uncomputable:
Sounds good to me.

delta1:
Done thanks @uncomputable @sanket1729 @apoelstra @stringhandler

PR #389, opened 2026-08-17
Relicence to apache or mit

stringhandler:
Relicence as suggested in #386 

This would need to be approved by sanket1729 and uncomputable as they are the current copyright holders. Alternatively we could set the date when the copyright switches over?

I didn't add the SPDX records as they are not 100% necessary and it's a lot of code changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "simplicityhl"
 version = "0.7.1"
 authors = ["sanket1729 <sanket1729@gmail.com>"]
-license = "CC0-1.0"
+license = "MIT OR Apache-2.0"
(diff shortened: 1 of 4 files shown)


sanket1729 (approved):

apoelstra (approved):
ACK dcdc19232f76a7b5e6faa523499d87613cefd42a; successfully ran local tests

KyrylR (approved):
utACK dcdc19232f76a7b5e6faa523499d87613cefd42a

delta1 (approved):
ACK dcdc19232f76a7b5e6faa523499d87613cefd42a; tested locally

PR #392, opened 2026-08-24
cli: add ability to set project root

KyrylR:
When files are not on the `project_root/simf` or `project_root/simf` or any other source dir but deeper, like `project_root/simf/math/module/*add*.simf` and in *.simf we specify the crate keyword it would resolve as a project_root `project_root/simf/math/module` but it should resolve as project_root/simf for the correct LSP analysis 

This PR introduces a way to provide a specific project root to help address the issue

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -70,6 +70,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                     .help("Link a dependency, optionally scoped to a specific module (e.g., --dep ./libs/merkle:math=./libs/math)"),
             )
             .arg(
+                Arg::new("project_root")
+                    .long("project-root")
+                    .value_name("PROJECT_ROOT")
+                    .action(ArgAction::Set)
+                    .help(
+                        "Project root for resolving `crate::` imports (defaults to the entry file's directory)",
+                    ),
+            )
+            .arg(
                 Arg::new("wit_file")
                     .long("wit")
                     .short('w')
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -69,6 +69,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                     .action(ArgAction::Append)
                     .help("Link a dependency, optionally scoped to a specific module (e.g., --dep ./libs/merkle:math=./libs/math)"),
             )
+            .arg(
+                Arg::new("project_root")
+                    .long("project-root")
+                    .value_name("PATH")
```
stringhandler:
```suggestion
                    .value_name("PROJECT_ROOT")
```

src/main.rs
```diff
@@ -157,11 +166,23 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    let canon_root = match matches.get_one::<String>("project_root") {
+        Some(project_root) => CanonPath::canonicalize(Path::new(project_root))?,
+        None => main_path
+            .as_path()
+            .parent()
+            .and_then(|p| CanonPath::canonicalize(p).ok())
+            .ok_or("Failed to determine project root directory from entry file")?,
+    };
+
+    if !main_path.starts_with(&canon_root) {
```
stringhandler, line 178: If this fails anyway, what is the purpose of allowing you to specify it? This feels like the behaviour is the same before this PR

stringhandler (commented):
ConceptACK, but I think you should allow the main entry path to exist outside of the project root.

LesterEvSe (approved):
ACK d21f237211ab1d0b4b4adfa5d8d91ef00771adc4; tested locally with `just check` and `just check_fuzz`

src/main.rs
KyrylR: Resolved by the discussion below

PR #394, opened 2026-08-24
Remove discouragement, add developer docs link

schoen:
Add link to development documentation and remove warning that says SimplicityHL is not yet ready to use.


stringhandler (commented):
Perhaps we should state the future versions of SimplicityHL may not compile to the same code as previous versions. Or at least "is not guaranteed to match previous versions"

delta1 (approved):

PR #395, opened 2026-08-24
Replace the deprecated relative distance timelock

schoen:
Switch the deprecated relative distance timelock call in the example contract examples/last_will.simf out, replacing it with a (proposed) stdlib function that enforces the same timelock policy correctly.


stringhandler (approved):
utACK eef3ece2b8d17d18b1c3259385a5f447e85f24d9

examples/last_will.simf
```diff
@@ -43,8 +43,9 @@ pub fn enforce_relative_distance(min_distance: Distance) {
 }
 
 fn inherit_spend(inheritor_sig: Signature) {
-    let days_180: Distance = 25920;
-    enforce_relative_distance(days_180);
+    // On Bitcoin, this Distance would be 180 days; on Liquid, it is 18 days.
+    let timelock_distance: Distance = 25920;
```
delta1, line 47: Since we're touching this and it's supposed to be for inheritance, what's the maximum possible value we could use here?

delta1: Maybe a nice to have in the future would be `Distance::MAX`

stringhandler: This is more like a dead man's switch where the funds unlock after 180 days of inactivity.

schoen: If someone wanted to actually use it for inheritance, the preferable thing would be to use `Duration`, which can be up to 65535 units of 512 seconds (about 1.06 years).

schoen:
@apoelstra Squashed!

apoelstra (approved):
ACK 1f769c9fe7d4f02bbd6346d98a3baa5a7c1ee17f; successfully ran local tests

KyrylR (approved):
ACK 5cdd0d8fd6016096b086eddb309a578ed7ed6371. Thanks for squashing the earlier commits. All local tests I ran passed: `cargo test --workspace --locked --quiet`.

LesterEvSe (approved):
ACK 5cdd0d8fd6016096b086eddb309a578ed7ed6371; tests locally with `just check` and `just check_fuzz`

PR #405, opened 2026-09-01
lint: enable clippy cast lints

stringhandler:
cast_possible_wrap and cast_sign_loss have no hits at all. Turn them on.

cast_possible_truncation had four hits, all provably safe by construction; annotate them at the site instead of allowing the lint crate-wide, so the existing 'every cast should contain a code comment' convention is enforced rather than aspirational.

LLM suggested and generated

### DIFF
--- a/codegen/src/main.rs
+++ b/codegen/src/main.rs
@@ -37,6 +37,9 @@ fn write_jet<W: io::Write>(jet: Elements, w: &mut W) -> io::Result<()> {
     write!(w, "pub fn {jet}(")?;
     let parameters = simplicityhl::jet::source_type(&jet);
     for (i, ty) in parameters.iter().enumerate() {
+        // Jets take a handful of parameters at most, so the index stays well
+        // inside `u8` (and inside the a-z range this names them from).
+        #[allow(clippy::cast_possible_truncation)]
         let identifier = (b'a' + i as u8) as char;
         if i == parameters.len() - 1 {
             write!(w, "{identifier}: {ty}")?;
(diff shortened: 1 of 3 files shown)


codegen/src/main.rs
```diff
@@ -37,6 +37,9 @@ fn write_jet<W: io::Write>(jet: Elements, w: &mut W) -> io::Result<()> {
     write!(w, "pub fn {jet}(")?;
     let parameters = simplicityhl::jet::source_type(&jet);
     for (i, ty) in parameters.iter().enumerate() {
+        // Jets take a handful of parameters at most, so the index stays well
+        // inside `u8` (and inside the a-z range this names them from).
+        #[allow(clippy::cast_possible_truncation)]
         let identifier = (b'a' + i as u8) as char;
```
apoelstra, line 43: In 724f851e3e6c7f11301b05429b62edcf7943ca3c:

We should use `u8::try_from` and `unwrap` if we really believe this is impossible.

src/num.rs
```diff
@@ -335,7 +343,11 @@ impl FromStr for U256 {
             // Add to the least significant bytes first
             for byte in bytes.iter_mut().rev() {
                 let value = u32::from(*byte) * 10 + carry;
-                *byte = (value % 256) as u8;
+                // A remainder modulo 256 is at most 255, so it fits in a `u8`.
+                #[allow(clippy::cast_possible_truncation)]
+                {
+                    *byte = (value % 256) as u8;
```
apoelstra, line 349: In 724f851e3e6c7f11301b05429b62edcf7943ca3c:

This is just confusing. We should write `*byte = value as u8` and whitelist the lint with "truncation is deliberate".

stringhandler: Yeah, I think let me fix these and then I'll undraft it

PR #407, opened 2026-09-01
refactor types module

apoelstra:
Just moves a bunch of types around. No behavioral changes. Should make the types module much easier to navigate.


delta1 (approved):
ACK eb4ad94f82bcf3ba9a6bf4eeeea2669743694e55; tested locally

src/ast.rs
```diff
@@ -1716,29 +1716,19 @@ fn analyze_enum_construction(
 /// conservative either way, since their partition layout complicates
 /// position alignment.
 fn cast_preserves_enum_identity(source: &ResolvedType, target: &ResolvedType) -> bool {
-    match (source.as_inner(), target.as_inner()) {
-        (TypeInner::Enum(src), TypeInner::Enum(dst)) => src == dst,
-        (TypeInner::Enum(_), _) | (_, TypeInner::Enum(_)) => false,
-        (TypeInner::Option(src), TypeInner::Option(dst)) => cast_preserves_enum_identity(src, dst),
-        (TypeInner::Either(src_l, src_r), TypeInner::Either(dst_l, dst_r)) => {
-            cast_preserves_enum_identity(src_l, dst_l) && cast_preserves_enum_identity(src_r, dst_r)
+    for (source, target) in source.post_order_iter().zip(target.post_order_iter()) {
```
stringhandler: In 7b31e94e0ca269484e61959e25947dfaa9694a17

LLM found this while I was reviewing with it. TL;DR: `zip` exits on the shorter of the two arrays, so potentially some lost data here.

Happy to provide the full output, but it was quite long. Not sure if there are cases where this might succeed if the shorter post_order_iter is a subset of the longer post_order_iter.

src/pattern.rs
```diff
@@ -51,25 +51,36 @@ impl Pattern {
...
+                    if let Some(types) = ty.as_tuple() {
+                        stack.extend(pats.iter().zip(types.iter().map(Arc::as_ref)));
+                    } else {
+                        return Err(Error::ExpressionUnexpectedType { ty: ty.clone() });
+                    }
                 }
-                (Pattern::Array(pats), TypeInner::Array(ty, size)) if pats.len() == *size => {
-                    stack.extend(pats.iter().zip(std::iter::repeat(ty.as_ref())));
+                Pattern::Array(pats) => {
+                    if let Some((ty, size)) = ty.as_array() {
```
stringhandler, line 75: In 7b31e94 LLM finding:
this line shadows the outer `ty` with the element type, so the length-mismatch arm reports the wrong type:


this PR: Expected expression of type `u8`; found something else
master:  Expected expression of type `[u8; 3]`; found something else
Renaming the binding to elem_ty fixes it. Minor: the commit message mentions as_list / lists, but Pattern has no List variant.

stringhandler (commented):
In 7b31e94e0ca269484e61959e25947dfaa9694a17

Happy with other previous commits, but seems like a regression introduced in the last one.

src/ast.rs
apoelstra: Ah, damn, you're right, this `zip`-based algorithm is just wrong. Let me rework it.

apoelstra: And I guess I should start LLM-reviewing my patches before I submit them..

src/pattern.rs
apoelstra: I replaced the explicit `return` value with an auxiliary closure which avoids the shadowing and also reduces code duplication.

Also updated the commit message.

src/ast.rs
apoelstra: The existing algorithm is also buggy (at least in the way mentioned in its doccomment). You can cast an `Option<u8>` to `Either<(), u8>` but not if you replace `u8` with an enum.

apoelstra: I've added several regression tests, including ones that fail on this code but not on the old code. I may need to just back out this change and move it to a separate PR because I think a correct algorithm is going to be a fair bit longer than this.

apoelstra: I just removed the change. But added a commit that adds regression test anyway.

stringhandler (approved):

PR #409, opened 2026-09-02
Eliminate quadratic environment rebuilds in Scope::get

tvolk131:
## What this fixes

Compiling a function with many `let` bindings was quadratic in the number of bindings. Every variable read rebuilt the entire input environment from scratch: it re-folded all live bindings into a tree, converted that tree, and then searched it for the variable.

## What changed

`Scope` now maintains its bindings incrementally: a flat list of binding patterns, a map from identifier to binding index, and per-scope start offsets. Reading a variable computes its take/drop selector directly from the binding's position: drop past the newer bindings, take into the target binding, then select within that binding's own pattern. The resulting selectors are identical to what the old code produced, so the compiled program is unchanged.

## Correctness

- The compiled program is unchanged. The pinned test-data bytecode regressions that CI already runs pass unmodified, and diffing the output of all example programs against a master build shows identical bytes throughout.
- A randomized property test compares the new selection logic against the old algorithm over random scope stacks, covering shadowing, nested scopes, function scopes, and compound patterns.

## Performance

Local benchmarking on an Apple M4 Pro, release build:

| program shape                        | codegen before | codegen after |
|--------------------------------------|---------------|---------------|
| 64 sequential bindings               | 3.0 ms        | 2.0 ms (-35%) |
| 256 sequential bindings              | 24.7 ms       | 7.9 ms (-68%) |
| 256 bindings across 32 nested scopes | 20.8 ms       | 7.7 ms (-63%) |
| 4096 sequential bindings             | 3.90 s        | 120 ms        |

Codegen speed is effectively completely linear with respect to the number of bindings now. A program with 4096 bindings (which is 64 * 64) takes 60x longer than one with 64 bindings.

## Commits

Two commits. The first is the optimization. The second is a pure refactor that moves the three coupled binding fields into a `ScopeBindings` type with its own invariant test. Happy to drop the second if you prefer the fields on `Scope`.

The benchmark and profiling harness behind these numbers lives on my fork ([bench/compiler-perf](https://github.com/tvolk131/simplicityhl/tree/bench/compiler-perf)). Happy to send it as a separate PR if useful.

### DIFF
--- a/src/compile/mod.rs
+++ b/src/compile/mod.rs
@@ -2,6 +2,7 @@
 
 mod builtins;
 
+use std::collections::HashMap;
 use std::sync::Arc;
 
 use either::Either;
@@ -16,9 +17,10 @@ use crate::ast::{
 };
 use crate::debug::CallTracker;
 use crate::error::{Diagnostic, Error, Span, WithSpan};
-use crate::named::{self, CoreExt, PairBuilder};
+use crate::named::{self, CoreExt, PairBuilder, SelectorBuilder};
 use crate::num::{NonZeroPow2Usize, Pow2Usize};
 use crate::pattern::{BasePattern, Pattern};
+use crate::str::Identifier;
 use crate::template_program::{TemplateProgram, TemplateProgramWitness};
 use crate::types::{StructuralType, TypeDeconstructible};
 use crate::value::{StructuralValue, Value};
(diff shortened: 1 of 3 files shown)


stringhandler (approved):
ACK f78cc90805f36c0755a4d5a7db1515f98a3be7e7 

Ran tests locally

PR #413, opened 2026-09-04
errors: dedup same compiler errors, present them in deterministic order

KyrylR:
The PR is meant to make the reported order of errors deterministic and deduplicated to improve UX

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,5 +1,5 @@
 use std::collections::hash_map::Entry;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::ffi::OsStr;
 use std::fmt;
 use std::io::{self, IsTerminal, Write};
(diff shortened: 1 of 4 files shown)


src/error.rs
```diff
@@ -397,9 +397,16 @@ where
...
     diags: Vec<Diagnostic>,
     error_count: usize,
+    index: Box<DiagnosticIndex>,
     sources: Option<SourceMap>,
 }
 
+#[derive(Debug, Clone, Default)]
+struct DiagnosticIndex {
+    identities: HashSet<String>,
+    source_order: HashMap<usize, usize>,
```
LesterEvSe, line 407: In this case `BTreeMap` will be more optimized than `HashMap`

```suggestion
    source_order: BTreeMap<usize, usize>,
```

apoelstra: I'm skeptical. We can't make use of the `BTreeMap`'s ordering since the actual presentation order depends on `source_rank` which uses data outside of the map. So it's a question of whether tree-rebalancing operations are cheaper than siphash operations.

If we are really concerned we can use a custom hasher which just casts the `usize`s to `u64`s, and that will definitely be faster than either approach.

In practice I think it doesn't matter at all, and `HashMap` (a) has a more full-featured API, and (b) communicates to the reader of the code that the order doesn't matter.

LesterEvSe: Okay then, let's just leave `HashMap`

LesterEvSe (approved):
ACK 5c8801015e40207cf08dbe1ff267e2db6edc6c08; tested locally with `just check` and `just check_fuzz`

PR #416, opened 2026-09-07
ast: make each use declaration atomic

KyrylR:
This makes each use declaration atomic: either every import succeeds, or the scope remains unchanged

Previously, `use source::{Good, Missing};` could insert `Good` before failing on `Missing`. The change stages imports temporarily and commits them only after the whole declaration succeeds

This prepares the compiler to recover from errors and report multiple diagnostics without leaked imports affecting later analysis


LesterEvSe (approved):
ACK 4a3cc9444031f7aba9efd19a40855659b67c461e; tested locally with `just check` and `just check_fuzz`

PR #417, opened 2026-09-17
Add uninhabited `Never` type

LesterEvSe:
## Description

Add the uninhabited type `Never` to `TypeInner`, together with `ResolvedType::never` and `ResolvedType::is_never`.

## Motivation

This is the first step of splitting #383. Today, analysis stops at the first error. To keep going, the compiler must give a broken expression *some* type. A placeholder such as `()` causes follow-up errors that are not real mistakes:

``` rust
let x: u32 = foo(;   // real error
let y: u32 = x + 1;  // with `()` as a placeholder: "expected u32, found ()"
```

If the broken expression has type `Never` instead, every later check that involves it passes silently, and each mistake is reported exactly once. The rule becomes simple: whoever creates a `Never` reports the error; whoever uses a `Never` stays silent.


apoelstra:
Nice. Straightforward enough, queued for ACK.

Right now we are using `!` as a placeholder for "code cannot be compiled" so panicking when we try to lower makes sense.

Eventually (not in this PR!) we'll want to add it to the language syntax and then we'll need to consider a few different cases and we'll need to adjust this lowering logic. In particular we likely will need to add a normalization step:

* Any product (or generalized product, e.g. `List`) with `!` can be promoted to `!`.
* Any sum or enum with `!` can have its `!` branch deleted; if we then have a single-arm sum it should be replaced with that arm.

And then after normalization, we need to consider what happens if we still try to lower a `!`:

* If `!` appears in the source type of an expression, that expression is uncallable and should not be compiled, so we can just throw it away. Rather than panicking we should return some sort of error indicating that we're lowering a type that cannot be constructed, and we should prune away whatever object it came from.
* If `!` appears in the *target* type of an expression, the implication is that the expression always diverges. Currently you cannot write such an expression, since it would need some sort of "unconditional panic" primitive. But we may want to add one. In this case, we should lower `!` to a free variable.

(I think that this "every HL type can be lowered to a Simplicity type" paradigm will eventually go away, but it's a very simple model so I'd like to keep it as long as we can.)

apoelstra (approved):
ACK 88070f98c0aed6772fd96cf0b0fe9f4a6a469eec; successfully ran local tests

LesterEvSe:
This is the first step of splitting #383. Today, analysis stops at the first error. To keep going, the compiler must give a broken expression *some* type. A placeholder such as `()` causes follow-up errors that are not real mistakes:

```
let x: u32 = foo(;   // real error
let y: u32 = x + 1;  // with `()` as a placeholder: "expected u32, found ()"
```

If the broken expression has type `Never` instead, every later check that involves it passes silently, and each mistake is reported exactly once. The rule becomes simple: whoever creates a `Never` reports the error; whoever uses a `Never` stays silent.

PR #418, opened 2026-09-18
AST: Return analysis errors through DiagnosticManager

LesterEvSe:
## Description

This PR prepares the AST phase for reporting multiple errors. `Program::analyze` now takes `&mut DiagnosticManager` and returns `Option<Self>` instead of `Result<Self, Diagnostic>`. Behaviour does not change: analysis still stops at the first error.

## Next steps

Follow-up PRs will extend analysis from one error to many, one level at a time:

1. Report every error inside an expression (tuple elements, call arguments, match arms), not only the first one. #420
2. Continue with the next item after an item fails.
3. Continue with the next statement after a statement fails.

Each step reports only independent errors, so it does not need poisoning to prevent cascading errors. Poisoning will come later, for the cases where analysis would otherwise stop, such as a broken declared type.


src/ast.rs
```diff
@@ -1352,37 +1362,60 @@ trait AbstractSyntaxTree: Sized {
...
+        };
+        let (parameters, witness_types, call_tracker, reported) = scope.destruct();
+
+        // The gate: analysis may continue after an error, but it never
+        // returns a program in that case, so such a program is never compiled.
+        if !reported.is_empty() {
+            diagnostics.extend(reported);
+            return None;
+        }
+        let main = main.expect("analysis fails only by reporting an error");
```
KyrylR: Weird new panic

LesterEvSe: Just got rid of it

KyrylR (approved):
ACK 1f279a4a6aface043ccf0c617600f49e200c1362; successfully ran local tests

PR #419, opened 2026-09-21
Fix typo in jet documentation (ration -> ratio)

schoen:


apoelstra (approved):
ACK 8dd86663b54a1332a39a6814419dac28f12b4955; successfully ran local tests

src/docs/jet.rs
```diff
@@ -375,7 +375,7 @@ Return `(u64::MAX, u64::MAX)` when the preconditions are not satisfied."#,
         Elements::GejDouble => "Double a point. If the result is the point at infinity, it is returned in canonical form.",
         Elements::GejEquiv => "Check if two points represent the same point.",
         Elements::GejGeAdd => "Add two points. If the result is the point at infinity, it is returned in canonical form.",
-        Elements::GejGeAddEx => "Add two points. Also return the ration of the `a`s z-coordinate and the result's z-coordinate. If the result is the point at infinity, it is returned in canonical form.",
+        Elements::GejGeAddEx => "Add two points. Also return the ratio of the `a`s z-coordinate and the result's z-coordinate. If the result is the point at infinity, it is returned in canonical form.",
```
KyrylR: AI finding: 

The returned field element is `rzr`, defined by `result.z = a.z * rzr` in the pinned `simplicity-sys` implementation. For a nonzero `a.z`, that is `result.z / a.z`. The current wording, 'ratio of the `a`s z-coordinate and the result's z-coordinate', can be read as the inverse. Could this specify the direction, for example, 'the ratio of the result's z-coordinate to `a`'s z-coordinate'? This wording predates the typo, but the edited description is still misleading.

KyrylR (commented):
All local tests passed: `cargo test --features docs` and `cargo test --workspace --all-features`. The normal all-features workspace docs build also succeeded.

src/docs/jet.rs
KyrylR: Will fix it myself

KyrylR (approved):
ACK a8ba74492385e860ac73b526f358fcf9f3f10217. The ratio-direction review comment is addressed. All test runs completed successfully: `cargo test --workspace --all-features --locked --quiet` on the exact commit passed 403 tests, with 5 existing ignores, and all 12 PR checks passed. `cargo doc --workspace --all-features --no-deps --locked` also succeeded.

LesterEvSe:
Checked this PR with Claude Opus 5. It agrees the change is correct, and suggested an expanded doc:

```
        Elements::GejGeAddEx => r#"Add the point `b`, in affine coordinates, to the point `a`, in Jacobian coordinates.

Return the z-coordinate ratio between the sum and `a`, followed by the sum itself. If `a` is not the point at infinity, then the ratio satisfies `sum.z = a.z * ratio`, which means `ratio = sum.z / a.z`.

- Return a ratio of `0` if `a` is the point at infinity; the sum is then `b`.
- If the sum is the point at infinity, it is returned in canonical form."#,
```

LesterEvSe (approved):
utACK a8ba74492385e860ac73b526f358fcf9f3f10217

PR #420, opened 2026-09-21
Report every error inside an expression

LesterEvSe:
Report every error inside an expression instead of only the first one:

- Analyze every element of tuples, arrays and lists, every call argument, both arms of a `match`, every arm of an enum `match` and every enum payload, even if some of them fail.
- Add tests.


KyrylR (approved):
ACK 12cc8faee9ca85384c6676898644790da44262db; successfully ran local tests

